PR #1, opened 2013-07-16
Added support for decryption of RAR files 

hrasyid:
As discussed in e-mail, here is the pull request which include RAR decryption. Please review and make any changes if necessary. Note that currently there is still problem due to inavailability of Cryptography in Portable.

### DIFF
--- a/SharpCompress/Common/Rar/RarCryptoWrapper.cs
+++ b/SharpCompress/Common/Rar/RarCryptoWrapper.cs
@@ -6,31 +6,31 @@ namespace SharpCompress.Common.Rar
 {
     internal class RarCryptoWrapper : Stream
     {
-        private readonly Stream _actualStream;
-        private byte[] _salt;
-        private RarRijndael _rijndael;
-        private readonly string _password;
-        private readonly Queue<byte> _data = new Queue<byte>();
+        private readonly Stream actualStream;
+        private byte[] salt;
+        private RarRijndael rijndael;
+        private readonly string password;
+        private readonly Queue<byte> data = new Queue<byte>();
 
         public RarCryptoWrapper(Stream actualStream, string password)
         {
-            _actualStream = actualStream;
-            _password = password;
+            this.actualStream = actualStream;
+            this.password = password;
         }
 
         internal byte[] Salt
         {
-            get { return _salt; }
+            get { return salt; }
             set
             {
-                _salt = value;
+                salt = value;
                 if (value != null) InitializeAes();
             }
         }
 
         private void InitializeAes()
         {
-            _rijndael = RarRijndael.InitializeFrom(_password, _salt);
+            rijndael = RarRijndael.InitializeFrom(password, salt);
         }
 
         public override void Flush()
@@ -50,13 +50,13 @@ namespace SharpCompress.Common.Rar
 
         public override int Read(byte[] buffer, int offset, int count)
         {
-            if (Salt == null) return _actualStream.Read(buffer, offset, count);
+            if (Salt == null) return actualStream.Read(buffer, offset, count);
             return ReadAndDecrypt(buffer, offset, count);
         }
 
         public int ReadAndDecrypt(byte[] buffer, int offset, int count)
         {
-            int queueSize = _data.Count;
+            int queueSize = data.Count;
             int sizeToRead = count - queueSize;
 
             if (sizeToRead > 0)
@@ -66,16 +66,16 @@ namespace SharpCompress.Common.Rar
                 {
                     //long ax = System.currentTimeMillis();
                     byte[] cipherText = new byte[RarRijndael.CryptoBlockSize];
-                    _actualStream.Read(cipherText, 0, RarRijndael.CryptoBlockSize);
+                    actualStream.Read(cipherText, 0, RarRijndael.CryptoBlockSize);
 
-                    var readBytes = _rijndael.ProcessBlock(cipherText);
+                    var readBytes = rijndael.ProcessBlock(cipherText);
                     foreach(var readByte in readBytes)
-                        _data.Enqueue(readByte);
+                        data.Enqueue(readByte);
 
                 }
 
                 for (int i = 0; i < count; i++)
-                    buffer[offset+i] = _data.Dequeue();
+                    buffer[offset+i] = data.Dequeue();
             }
             return count;
         }
@@ -109,7 +109,7 @@ namespace SharpCompress.Common.Rar
 
         protected override void Dispose(bool disposing)
         {
-            if(_rijndael!= null) _rijndael.Dispose();
+            if(rijndael != null) rijndael.Dispose();
             base.Dispose(disposing);
         }
     }
--- a/SharpCompress/Common/Rar/RarRijndael.cs
+++ b/SharpCompress/Common/Rar/RarRijndael.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
@@ -28,10 +27,10 @@ namespace SharpCompress.Common.Rar
 
             rijndael = new RijndaelManaged() { Padding = PaddingMode.None };
             aesInitializationVector = new byte[CryptoBlockSize];
-            int rawLength = 2 * this.password.Length;
+            int rawLength = 2 * password.Length;
             byte[] rawPassword = new byte[rawLength + 8];
-            byte[] passwordBytes = Encoding.UTF8.GetBytes(this.password);
-            for (int i = 0; i < this.password.Length; i++)
+            byte[] passwordBytes = Encoding.UTF8.GetBytes(password);
+            for (int i = 0; i < password.Length; i++)
             {
                 rawPassword[i * 2] = passwordBytes[i];
                 rawPassword[i * 2 + 1] = 0;


SharpCompress.Test/SharpCompress.Test.csproj
```diff
@@ -42,7 +42,7 @@
     </StartupObject>
   </PropertyGroup>
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
+    <SignAssembly>false</SignAssembly>
```
adamhathcock: Revert this please.

SharpCompress/AssemblyInfo.cs
```diff
@@ -14,7 +14,7 @@
 [assembly: AssemblyProduct("SharpCompress")]
 [assembly:
     InternalsVisibleTo(
-        "SharpCompress.Test, PublicKey=002400000480000094000000060200000024000052534131000400000100010005d6ae1b0f6875393da83c920a5b9408f5191aaf4e8b3c2c476ad2a11f5041ecae84ce9298bc4c203637e2fd3a80ad5378a9fa8da1363e98cea45c73969198a4b64510927c910001491cebbadf597b22448ad103b0a4007e339faf8fe8665dcdb70d65b27ac05b1977c0655fad06b372b820ecbdccf10a0f214fee0986dfeded"
```
adamhathcock, line 17: Revert this please.

SharpCompress/Common/Rar/Headers/RarHeaderFactory.cs
```diff
@@ -9,22 +10,27 @@ internal class RarHeaderFactory
...
-        internal RarHeaderFactory(StreamingMode mode, Options options)
+        internal RarHeaderFactory(StreamingMode mode, Options options, string password = null)
         {
             StreamingMode = mode;
             Options = options;
+            Password = password;
         }
 
         private Options Options { get; set; }
+        public string Password { get; set; }
```
adamhathcock: private setter?

SharpCompress/Common/Rar/Headers/RarHeaderFactory.cs
```diff
@@ -164,7 +182,7 @@ private RarHeader ReadNextHeader(Stream stream)
                                 {
                                     ReadOnlySubStream ms
                                         = new ReadOnlySubStream(reader.BaseStream, fh.CompressedSize);
-                                    fh.PackedStream = ms;
+                                    fh.PackedStream = new RarCryptoWrapper(ms, Password) { Salt = fh.Salt};
```
adamhathcock: Shouldn't this be only wrapped if it is encrypted?

SharpCompress/Common/Rar/RarCryptoWrapper.cs
```diff
@@ -0,0 +1,116 @@
...
+                int alignedSize = sizeToRead + ((~sizeToRead + 1) & 0xf);
+                for (int i = 0; i < alignedSize/16; i++)
+                {
+                    //long ax = System.currentTimeMillis();
+                    byte[] cipherText = new byte[RarRijndael.CryptoBlockSize];
+                    _actualStream.Read(cipherText, 0, RarRijndael.CryptoBlockSize);
+
+                    var readBytes = _rijndael.ProcessBlock(cipherText);
+                    foreach(var readByte in readBytes)
+                        _data.Enqueue(readByte);
```
adamhathcock: Would using a queue like this be a source of performance problems?  Also, please nest everything with braces.  Even if it's just a single line under an foreach for if statement.

SharpCompress/Common/Rar/RarRijndael.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace SharpCompress.Common.Rar
+{
+    class RarRijndael : IDisposable
```
adamhathcock: Always have visibility modifiers.

SharpCompress/Common/Rar/RarRijndael.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private readonly string _password;
```
adamhathcock: I don't use underscores for field variables.  It's cleaner and Resharper colors them differently.  I really recommend Resharper :)

SharpCompress/IO/MarkingBinaryReader.cs
```diff
@@ -65,45 +126,49 @@ public override char[] ReadChars(int count)
...
         public override decimal ReadDecimal()
         {
-            CurrentReadByteCount += 16;
-            return base.ReadDecimal();
+            return ByteArrayToDecimal(ReadBytes(16), 0);
+        }
+
+        private decimal ByteArrayToDecimal(byte[] src, int offset)
+        {
+            //http://stackoverflow.com/a/16984356/385387
```
adamhathcock, line 134: Thanks for changing this.  The link is very useful.

SharpCompress/SharpCompress.Portable.csproj
```diff
@@ -35,10 +35,11 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
+    <SignAssembly>false</SignAssembly>
   </PropertyGroup>
   <PropertyGroup>
-    <AssemblyOriginatorKeyFile>SharpCompress.pfx</AssemblyOriginatorKeyFile>
```
adamhathcock, line 41: Revert please.

SharpCompress/SharpCompress.Portable.csproj
```diff
@@ -35,10 +35,11 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
```
adamhathcock, line 38: Revert please.

SharpCompress/SharpCompress.Portable.csproj
```diff
@@ -286,9 +288,6 @@
     <Compile Include="Writer\Zip\ZipCentralDirectoryEntry.cs" />
     <Compile Include="Writer\Zip\ZipWriter.cs" />
   </ItemGroup>
-  <ItemGroup>
```
adamhathcock, line 289: Revert please.

SharpCompress/SharpCompress.csproj
```diff
@@ -73,10 +73,11 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
+    <SignAssembly>false</SignAssembly>
   </PropertyGroup>
   <PropertyGroup>
-    <AssemblyOriginatorKeyFile>SharpCompress.pfx</AssemblyOriginatorKeyFile>
```
adamhathcock, line 79: Revert please.

SharpCompress/SharpCompress.csproj
```diff
@@ -73,10 +73,11 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
```
adamhathcock, line 76: Revert please.

SharpCompress/SharpCompress.csproj
```diff
@@ -367,9 +370,6 @@
       <Install>true</Install>
     </BootstrapperPackage>
   </ItemGroup>
-  <ItemGroup>
```
adamhathcock, line 370: Revert please.

SharpCompress/Common/Rar/RarRijndael.cs
hrasyid: I see.. Interestingly, the default rule in my Resharper does the opposite (always recommend underscore for fields and colour the code if it doesn't have one). In fact my habit switched from no underscore to underscore because of Resharper and I think it's also more useful in avoiding name clash between field and method parameter :)

If the code consistently has no-underscore behaviour, I'll change it and change my resharper rule for this project.

adamhathcock: I should actually add my project level resharper settings then you'll just get them :)

SharpCompress/Common/Rar/RarCryptoWrapper.cs
hrasyid: The queue is necessary because we can only decrypt 16 bytes at a time, but the client of this code does not necessarily read 16 block at a time. As a solution for this, all 16 bytes will go to a queue, and if for instance the client only reads 2 bytes, the other 14 will stay in the queue and will be directly read (without decrypting) the next time.

Also, performance wise I don't think there's an issue, because generarlly, queues only involve in-memory access without any difficult calculation. I checked, the performance culprit in this class is the SHA1 calculation.

hrasyid: About braces, don't additional braces make the code look more verbose? If you're worried that people might add another line and forget to use braces, I think with IDEs like Visual studio it won't happen because the indentation will make it look obvious

adamhathcock: It does add more lines yes.  However, there are two reasons I do it.  First, it just makes it perfectly clear on a quick skim of the code that a line is nested.  Second, and probably more important, merging goes a lot smoother with the nesting.  I've shot myself in the foot with merging because the tool couldn't see the differences with the single line versus the nested approach.

Having a couple of extra lines because of braces never hurt anyone :) Of course, you can write some nasty code because of too many levels of nesting in a method, but that's a different story.

hrasyid: ok.. updated to use braces

PR #40, opened 2015-01-14
Inconsistent time format

Strachu:
Currently, not all archive types return the modification time of entry in the same format. ZIP returns the time in local time, while TAR and 7Z returns UTC time.
This has caused that in my timezone the modification time for .zip entries were correct, while the ones for .tar and .7z entries were 1 hour off.

I have fixed it, so that now every archive returns time as local time.

I was not sure whether it's better to always return local or UTC time, but I have decided to return local time as it is a bit more intuitive for a user. In the long term, it does not matter, as there is conversion available between them with the usage of .ToLocalTime() and ToUniversalTime() .methods, what matters is that returned times have to be consistent for all supported formats.


adamhathcock:
I would have said UTC would be better but it is just a matter of preference that way.  Consistency is the best.  Thanks!

PR #44, opened 2015-03-13
Added IWritableArchive interface

catester:
So that when working with ArchiveFactory, you can cast the opened
archive to IWritableArchive for accessing AddEntry, RemoveEntry and
SaveTo methods without having to know what type of archive is opened
underhood. Also moved extension methods from AbstractWritableArchive to
this new interface. Also added GZipArchive to ArchiveFactory.Create as
it is one of the 3 writable archive types (zip, tar, gzip), not sure why
it was omitted.

Also replaced NotImplementedException for streams.
Especially for streams, it is more appropriate to throw
NotSupportedException instead of NotImplementedException. Usually
consumers of streams expect NotSupportedException to handle errors.
For reference about this best practice, please see
these articles:
http://blogs.msdn.com/b/brada/archive/2004/07/29/201354.aspx
http://blogs.msdn.com/b/jaredpar/archive/2008/12/12/notimplementedexception-vs-notsupportedexception.aspx


adamhathcock:
Thanks for the clean up!

PR #52, opened 2015-04-07
Tar long name support

sander2:
Support for writing arbitrarily long filenames in tar archives.

The existing code used the ustar prefix field for supporting filepaths up to 255 chars (the actual maximum filepath length depends on the exact location of the slashes in the pat). However, there was a bug (several bugs, in fact) in the implementation that caused the filewriter to throw an error when a filepath of > 100 chars was used.
This pull request drops the use of the ustar prefix field in favor of the use of LongLinks. As far as I can tell, the use of LongLink is more common than the use of the ustar prefix field. Both 7zip and gnu tar use LongLinks by default. 
Since the previous implementation never worked anyway, there are no users depending on the use of the ustar prefix field.

### DIFF
--- a/SharpCompress.Test/Tar/TarArchiveTests.cs
+++ b/SharpCompress.Test/Tar/TarArchiveTests.cs
@@ -45,6 +45,44 @@ namespace SharpCompress.Test
         }
 
         [TestMethod]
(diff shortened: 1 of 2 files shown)


adamhathcock:
Thanks!

PR #55, opened 2015-05-08
Added read support for RARs with Protect Headers

benshoof:
Some RARs with recovery records contain Protect Headers, I've added
support for parsing them so that RARs containing them can be read,
instead of an invalid-header exception being thrown. Parsing logic taken
from unrar reference source.


adamhathcock:
Thanks!

PR #56, opened 2015-05-17
Improved testability

KOLANICH:
1) disabled requirement of signature
2) added runtime discovery of folder with archives

changes by this commit are already present in 13ef1c3, but they weren't merged into master

### DIFF
--- a/SharpCompress.Test/SharpCompress.Test.csproj
+++ b/SharpCompress.Test/SharpCompress.Test.csproj
@@ -42,7 +42,7 @@
     </StartupObject>
   </PropertyGroup>
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
+    <SignAssembly>false</SignAssembly>
   </PropertyGroup>
   <PropertyGroup>
     <AssemblyOriginatorKeyFile>..\SharpCompress\SharpCompress.pfx</AssemblyOriginatorKeyFile>
--- a/SharpCompress.Test/TestBase.cs
+++ b/SharpCompress.Test/TestBase.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.IO;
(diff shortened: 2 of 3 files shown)


adamhathcock:
If the internals visible to isn't necessary anymore, good!  I didn't notice.

If that path thing works, I don't see why not.  Saves headaches

@norvegec Everything is MSTest specific.  I think the TestContext only gets filled in per test method.  It might be null when ClassInitialize gets called but I'll let @KOLANICH say so or not.

KOLANICH:
>And [TestInitialize] is per test method rather than [ClassInitialize] or original static field initializers.
TestContext is not accessible when [ClassInitialize] is triggered. It is not static.

norvegec:
@KOLANICH actually, you can use TestContext like this:
```c#
       [ClassInitialize]static void ClassSetup(TestContext ctx)
       {
            TEST_BASE_PATH = Path.GetDirectoryName(Path.GetDirectoryName(ctx.TestDir));
            // init other static vars
       }
```
The problem here is the **lack** of  inheritance (WOW!), that makes this nice feature completely useless in our case right now.

KOLANICH:
@norvegec, thanks, learned it.

norvegec:
I suggest to replace 2) with the following:
```c#
[TestClass] // otherwise all derived will ignore AssemblyInitialize-marked method
    public class TestBase
    {
        [AssemblyInitialize] // runs only once for whole assembly
        public static void AssemblySetup(TestContext ctx)
        {
            TEST_BASE_PATH = Path.GetDirectoryName(Path.GetDirectoryName(ctx.TestDir));
            //init others
         }
```
The benefits here are a)speed improvements b)encapsulation of init logic c)if test are run in-parallel we ppbly don't need to invent the wheel with monitor/lock solution

KOLANICH:
Thank you. Fixed. Tests now are really going faster.

PR #63, opened 2015-06-11
ZipWriter: use ArchiveEncoding.Default

mrgleba:
ShaprCompress is used internally in Mono to implement
System.IO.Compression, which allows the user to specify path encoding.
The change allows for specifying the encoding via
ArchiveEncoding.Default.


adamhathcock:
Thanks for the fix!

PR #69, opened 2015-07-15
Skip entry stream on dispose

pnewman8:
Until now the caller had to completely consume each entry stream, or call SkipEntry(), before disposing the stream. If not, exception was thrown: "EntryStream has not been fully consumed". Hugely inconvenient; a user-thrown exception inside a "using (EntryStream)" block would be discarded.

Now automatically skips the entry on dispose.

Added method EntryStream.Cancel(). Call this if entry stream is unfinished, and no further entries are required. Helps with efficiency, as it avoids reading data that is not needed.

### DIFF
--- a/SharpCompress/Common/EntryStream.cs
+++ b/SharpCompress/Common/EntryStream.cs
@@ -1,16 +1,19 @@
 ﻿using System;
 using System.IO;
+using SharpCompress.Reader;
 
 namespace SharpCompress.Common
 {
     public class EntryStream : Stream
     {
+        public IReader Reader { get; private set; }
         private Stream stream;
         private bool completed;
         private bool isDisposed;
 
-        internal EntryStream(Stream stream)
+        internal EntryStream(IReader reader, Stream stream)
         {
+            this.Reader = reader;
             this.stream = stream;
         }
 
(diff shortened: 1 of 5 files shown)


SharpCompress/Common/EntryStream.cs
```diff
@@ -26,12 +26,27 @@ public void SkipEntry()
...
         }
 
+        public bool Cancelled { get; private set; }
+
+        /// <summary>
+        /// Indicates that the remainder of the stream is not required.
+        /// On dispose, the entry will not be skipped, so it helps with efficiency.
+        /// The downside is that subsequent entries are not usable, as the compressed stream is not positioned at an entry boundary.
+        /// </summary>
+        public void Cancel()
```
adamhathcock: I don't like this because it does invalidate the Reader altogether.  If you really want to abort reading entries, you should abort the Reader itself.  Not the stream for a single entry in the Reader.

pnewman8:
That's always been the case, simply by not consuming all the content. Cancelling doesn't change that.

I suppose the EntryStream could retain a reference to the reader, if you prefer. The crucial thing is that the EntryStream must know whether cancelled, on dispose.

Keep in mind that the consumer of the stream doesn't know what sort of stream it is, and doesn't know what a reader is. Default behaviour should therefore be to skip, with no exception.

adamhathcock:
I think it makes sense for the EntryStream to know it's Reader then a user can do Stream.Reader.Cancel() or something to abort everything.  Then the EntryStream can Dispose properly.

I guess I think that putting Cancel on the stream will lead to more exceptions with weird usage anyway.

pnewman8:
Fair enough. I will do another commit a bit later. Thanks Adam.

adamhathcock:
Thanks, this looks a lot cleaner to me.

PR #72, opened 2015-07-26
Extract Options And Total Sizes

hodm:
Fixed TotalSize For 7z
added TotalUncompressSize Tested for 7z
this enables to show progress for the entire archive
Added 2 Extract Options: PreserveFileTime And PreserveAttributes.
Put All the Log Command under DEBUG Condition.


adamhathcock:
This looks great.  Thanks!

adamhathcock:
The time comparison tests fail.  Not because the times are wrong but because it's comparing two different times because the Last Write values aren't the same.

PR #81, opened 2015-08-03
Move closing parentheses to where they should be, so build does not b…

haykpetros:
…reak for non-DEBUG configurations.


adamhathcock:
Thanks!

PR #82, opened 2015-08-04
Issue 79

haykpetros:
1. Removed one unused method.
2. Removed conditional compilation so now only managed version of WriteLittleEndian method will be used both for regular as well as Portable versions of code.


adamhathcock:
Did I ask why did you want to have no unsafe code?  I'm running on little sleep so forgive me.

haykpetros:
Well, I see no reason to have a method with conditional compilation that actually does same thing. I doubt there will be performance gains, but you always can run into memory leak when using pointers. Not to mention that they are harder to understand for newer developers I believe.

PR #83, opened 2015-08-05
Added additional check to make sure that data is properly copied to a…

haykpetros:
…rray regardless fo computer/CPU platform (little-endian or big-endian). In case of big-endian platform intermediate array will be reversed prior to copying to destination array.

### DIFF
--- a/SharpCompress/Utility.cs
+++ b/SharpCompress/Utility.cs
@@ -202,6 +202,10 @@ namespace SharpCompress
         public static void WriteLittleEndian(byte[] array, int pos, short value)
         {
             byte[] newBytes = BitConverter.GetBytes(value);
+
+            if (!BitConverter.IsLittleEndian)
+                Array.Reverse(newBytes);
+
             Array.Copy(newBytes, 0, array, pos, newBytes.Length);
         }
 
@@ -234,6 +238,10 @@ namespace SharpCompress
         public static void WriteLittleEndian(byte[] array, int pos, int value)
         {
             byte[] newBytes = BitConverter.GetBytes(value);
+
+            if (!BitConverter.IsLittleEndian)
+                Array.Reverse(newBytes);
+
             Array.Copy(newBytes, 0, array, pos, newBytes.Length);
         }
 


adamhathcock:
The algorithms want to run in little endian explicitly, regardless of platform.  We don't want to change endianness of the bytes because of platform for compression/decompression

mlischka:
I think `BitConverter.GetBytes` splits the short/int to bytes despite of the endianess and does **not** change the byte order. That means, when you are on a BigEndian system, you have to reverse the order to ensure the bytes are writtem in little-endian order to target byte-array; that's what haykpetros is doing. In my view his code is right.

adamhathcock:
I get what you're saying but if you look at the usage of the WriteLittleEndian method then you can see its used within the context of the VM of the algorithm which expects bytes to be moved Ina certain way.

There could still be problems with endianness when reading or writing outside the context of the algorithm but inside it should be just one way.

I guess I'd like to see tests for a big endian platform before pulling this in because I don't think it makes sense with my current understanding.

adamhathcock:
Oh wait....I just now got it.

I'm still hesitant to merge this in with out some big endian platform testing.  I mean, I didn't write this code originally and don't pretend to fully understand how it's supposed to work on all platforms.

haykpetros:
I believe it will be safe to merge my pull request as MSDN documentation is very clear on how exactly BitConverter.GetBytes() method works:

> The order of bytes in the array returned by the GetBytes method depends on whether the computer architecture is little-endian or big-endian.

So that is why adding BitConverter.IsLittleEndian check will resolve issue related to Big-endian systems:

> This value is true if the architecture is little-endian; false if it is big-endian.

> Different computer architectures store data using different byte orders. "Big-endian" means the most significant byte is on the left end of a word. "Little-endian" means the most significant byte is on the right end of a word.

Following two links clearly explain everything:

https://msdn.microsoft.com/en-us/library/system.bitconverter(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/de8fssa4(v=vs.110).aspx

PR #97, opened 2015-10-01
Add explicit compressioninfo when writing file to zip

nadyaA:
We use the ZipWriter to create archive with compression type deflate. But there is a problem - the default Mac UI unzipping (double click on the zip file) is throwing "Unable to expand. Error 2 No such file or directory." if there is an empty file in the archive. It only works  for archives that store the empty files and do not compress them.

As this is not a general bug, but one tool issue, the solution for me is to provide the API with option to explicitly change the compression info for specific files. This way if there are empty files or other specifics one can have different compression infos in one archive.


adamhathcock:
Interesting.  I like this.  Thanks.

Issue #93, opened 2015-09-14
ZipArchive.Open a zip file(the zipfile has my head flag) cann't read ZipArchiveEntry 


andyhebear:
//
            //read my zip file with head
            //
             using (FileStream fs = new FileStream("mypack.data.zip", FileMode.Open, FileAccess.Read, FileShare.Read)) {
                 byte[] buf = new byte[1024];
                 int offset=fs.Read(buf, 0, buf.Length);
                 System.Diagnostics.Debug.Assert(offset==1024);
                 ZipArchive zip=ZipArchive.Open(fs, Options.LookForHeader);//cann't read
                 //ZipArchive zip = ZipArchive.Open(fs, Options.None); //will throw exption
                 //ZipArchive zip = ZipArchive.Open(fs, Options.KeepStreamsOpen);//cann't read
                 foreach (ZipArchiveEntry zf in zip.Entries) {
                     Console.WriteLine(zf.Key);
                     //bug:the will not none in zipfile
                 }
             }
        }
     
you can find the code in
 https://github.com/RainsSoft/sharpcompress/blob/master/SharpCompress/_TestSharpCompress.cs
the "mypack.data.zip" you can open with winrar.exe,but cann't read in this code
https://github.com/RainsSoft/sharpcompress/blob/master/SharpCompress/mypack.data.zip

PR #98, opened 2015-10-13
read zipfile data with a offset position has bug

andyhebear:
this is a test for  read zipfile data with a offset position


adamhathcock:
Are you readding support for .NET 2.0?

You also need to update from master.

andyhebear:
base on .net3.5 ,i want use this lib in unity3d 5

andyhebear:
can you see the problem for #93

adamhathcock:
What are all these meta files?

andyhebear:
oh... this some unity3d project auto create files.

PR #99, opened 2015-10-27
Add password when opening RarReader

Rovak:
When using the following code the program will crash because there is no password specified

```c#
var archive = RarArchive.Open(***);
var entries = archive.ExtractAllEntries();
while(entries.MoveToNextEntry()) { // cryptographic exception here
   // do something with entries
}
```

This pull requests passes the password to the `RarReader`


adamhathcock:
Oh I see.  If the Archive is SOLID and Password protected, it threw an exception?

PR #101, opened 2015-11-19
Non-compilable statements + zip archive handling  

twirpx:
- Fixed non-compilable statements in DelfateManager
- Eliminated "throw - catch all" logic in ZipArchive (performance issue)
- ZipArchive with comments handling


adamhathcock:
Thanks!

Issue #79, opened 2015-08-03
Compiler Warning (level 3) CS0675 in DefalteManager.cs


haykpetros:
When building SharpCompress project with Warning level above 2 (usually 4 is default) it will fail on lines 683 and 695 of DeflateManager.cs:

```cs
unchecked
            {
                if (bi_valid > (int) Buf_size - len)
                {
                    //int val = value;
                    //      bi_buf |= (val << bi_valid);

                    bi_buf |= (short) ((value << bi_valid) & 0xffff); <-- THIS IS WHERE IT FAILS
                    //put_short(bi_buf);
                    pending[pendingCount++] = (byte) bi_buf;
                    pending[pendingCount++] = (byte) (bi_buf >> 8);


                    bi_buf = (short) ((uint) value >> (Buf_size - bi_valid));
                    bi_valid += len - Buf_size;
                }
                else
                {
                    //      bi_buf |= (value) << bi_valid;
                    bi_buf |= (short) ((value << bi_valid) & 0xffff); <-- THIS IS WHERE IT FAILS
                    bi_valid += len;
                }
            }
```
I believe following line:
```cs
bi_buf |= (short) ((value << bi_valid) & 0xffff);
```
can be replaced using following code instead:
```cs
bi_buf = (short)(bi_buf | (short)((value << bi_valid) & 0xffff));
```

adamhathcock:
Thanks for this, could you make this a Pull Request by chance?

haykpetros:
It seems that this might be a bug in Visual Studio 2015, so we are going to wait until it is verified.

Gachl:
Considering that there's a [wdc in MSDN about this](https://msdn.microsoft.com/en-us/library/wdc6717a.aspx) I doubt this is a bug.

benshoof:
It's never a bug in the compiler... except this time. Even worse, a commit has been made to SharpCompress that tried to fix this but instead broke the build on VS2013 with a legitimate case of the warning it tried to prevent.

The deflate code is from DotNetZip from 2009 and has been working without any compiler saying peep until VS2015 showed up and threw CS0675. This change in compiler behavior is a bug: https://github.com/dotnet/roslyn/issues/4027

[...]

PR #104, opened 2015-12-15
Fix VS2013 compiler warnings (errors)

benshoof:
Fixes broken build in VS2013 introduced by
18bd8102281c04c7eb09ecfdd984d7382b629f2c. That commit attempted to fix a
compiler warning from VS2015, but this turns out to be a compiler bug:
https://github.com/dotnet/roslyn/issues/4027 . That commit added code
which VS2013 correctly treats as a compiler warning, breaking the VS2013
build.
I have reverted this unnecessary change to the deflate code, fixing the
VS2013 build, and disabled warning CS0675 on send_bits() which will
satisfy VS2015.

Fixes https://github.com/adamhathcock/sharpcompress/issues/79


adamhathcock:
This is better.

PR #105, opened 2015-12-15
Fix Release build of Tests

benshoof:
Fixes release builds of SharpCompress.Test and SharpCompress.Test.Portable. The UNSIGNED symbol was missing from the Release configurations of SharpCompress.Unsigned and SharpCompress.PortableTest

### DIFF
--- a/SharpCompress/SharpCompress.PortableTest.csproj
+++ b/SharpCompress/SharpCompress.PortableTest.csproj
@@ -63,7 +63,7 @@
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
     <OutputPath>..\bin\</OutputPath>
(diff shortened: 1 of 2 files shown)


adamhathcock:
Ugh, I hate strong naming.  I was hoping .NET Core would free us from this.

benshoof:
It is a pain. In this case it's just the Debug and Release symbols that are out of sync which can easily happen when you have different ones for each configuration. At this point the only difference is DISABLE_TRACE which is defined on the Debug configurations but doesn't do anything. If DISABLE_TRACE is removed then the symbols will be the same for all configurations and the "All Configurations" mode of the VS project page can be used, that might discourage these getting out of sync in the future.

PR #107, opened 2015-12-15
Fixed serious performance regression (revert 0f12a073af4486509dfd777da1854efb5dc571ee)

benshoof:
Revert commit that caused all non-zip files to be read entirely upon opening.
IsZipArchive() would read and process the entire file looking for a zip header.


aemengo:
Nice catch, just ran into this myself.

adamhathcock:
Finally did a release for this

PR #113, opened 2016-01-04
Preserve File Timestamps with IReader

zentron:
Previously only `IArchive` would use the `ExtractionOptions` enum to set the timestamps & attributes on extracted files. With this change the common logic is extracted out and used during both `IReader` and `IArchive` extraction.


adamhathcock:
Thanks!

PR #114, opened 2016-01-07
fix:extract rar file incorrectly in linux platform because of "\\"

GWIian:
just debug in linux and find this problem.

### DIFF
--- a/SharpCompress/Common/Rar/Headers/FileHeader.cs
+++ b/SharpCompress/Common/Rar/Headers/FileHeader.cs
@@ -4,230 +4,234 @@ using SharpCompress.IO;
 
 namespace SharpCompress.Common.Rar.Headers
 {
-    internal class FileHeader : RarHeader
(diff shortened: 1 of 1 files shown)


adamhathcock:
Can't read the diff because the line endings changed :(

For better or worse, the line endings are windows.  I ought to change that.

adamhathcock:
I've copy/pasted from your repo but I still don't understand.  You're going to have to explain the issue.

libConch:
The problem seems will occure while extract rar-package on linux.

Build a rar package with only one file [AAAA\BBBB.txt] on windows, Then the path in rar-package is "AAAA\BBBB.txt". But if you decompress it on linux, the '\\' will not changed to '/', so error occure.

adamhathcock:
Making this a duplicate of https://github.com/adamhathcock/sharpcompress/pull/238

Line endings were hard to read

PR #115, opened 2016-01-17
Adds Non Lineal "Solid Rar" Extraction to RarArchive.

maxpiva:


adamhathcock:
I guess if people want this functionality it would have to be done something like this way.  I may do some clean up later between this and other solid functions.

Thanks!

maxpiva:
You welcome :)

Regards

adamhathcock:
Actually, I just undid this.  Sorry.

If the archive is SOLID, you should use the ExtractAllEntries which gives you an IReader interface that does this.

PR #118, opened 2016-01-22
	Bugfix for inflate algorithm cutting of the end of certain blocks

kenkendk:
After tracking down a case of a corrupt file being extracted, I was able to isolate the issue to a single 100kb file. If I compress this file with the maximum compression, and decompress it, I the loose the trailing 109 bytes.

I have finally tracked down the issue and made this patch. The problem is that the number of bits requested is larger than the number of bits required. For the `LEN` and `DIST` case, the number of required bits `k`  is set to `lbits` and `dbits` respectively. But if you look in the code below, you can see that they "consume" a variable number of bits, which in this case is less than `k`.

For most cases this is not a problem, but when the last code to be processed hits this case, the decompression stops prematurely and this cuts of a number of bytes. Other zip implementations (unarchiver, 7zip, etc) handles this correctly. This bug is also present in jzlib and the DotNetZip libraries, but not in the SharpZipLib/ICSharp.ZipLib.

If have a file that can reproduce this, but I am awaiting clarification to see if I can provide publicly it.

As a side note, the inflater file contains some of the worst code I have seen. It is a mix of states stored in objects and scope, copied around. There are almost no comments, tens of single digit variables, and the control flow is ridiculously hard to follow. In many places the same code is simply copy-pasted rather than being encapsulated, which is why there are two changes to the source code. In other words: if you have another zip implementation, I suggest you go with that :).

### DIFF
--- a/SharpCompress/Compressor/Deflate/Inflate.cs
+++ b/SharpCompress/Compressor/Deflate/Inflate.cs
@@ -895,11 +895,11 @@ namespace SharpCompress.Compressor.Deflate
                                 r = ZlibConstants.Z_OK;
                             else
                             {
-								// Handling missing trailing bit(s)
-								var tmp_tindex = (tree_index + (b & InternalInflateConstants.InflateMask[k]))*3;
-								if (k >= tree[tmp_tindex + 1])
-									break;
-								
+                                // Handling missing trailing bit(s)
+                                var tmp_tindex = (tree_index + (b & InternalInflateConstants.InflateMask[k]))*3;
+                                if (k >= tree[tmp_tindex + 1])
+                                    break;
+                                
                                 blocks.bitb = b;
                                 blocks.bitk = k;
                                 z.AvailableBytesIn = n;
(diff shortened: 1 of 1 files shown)


adamhathcock:
This implementation of inflate is from an older version of DotNetZip.  I have done very little to modify it.  So I'm not surprised DotNetZip actually has this issue too.

I would rather use another implementation.  In fact, I may use the new .NET Core's implementation.  I'd rather reference but I may have to C/P it as it's not publicly exposed.

Thanks for tracking this down.  I'm sure the DotNetZip guys would like to know about this.

kenkendk:
@adamhathcock: I saw you issue about using .Net Core and I think that would be better.

The test file which exhibits this behavior can be found here:
https://dl.dropboxusercontent.com/u/15251537/inflate-bug-test-stream.zip

Last commit on the DotNetZip codeplex site is from 2012:
https://dotnetzip.codeplex.com/SourceControl/list/changesets

It appears they have rolled it all up and are doing some commercial library instead.
I have notified the repo that is listed on NuGet for DotNetZip:
https://github.com/haf/DotNetZip.Semverd/issues/61

haf:
DotNetZip is not a commercial lib however, and it's based only around people doing PRs to the repository.

kenkendk:
@haf My bad, I read on the codeplex site (that appears abandoned) that Xceed was hosting it, and I could not find any sourcecode there, so I assumed they made DotNetZip a commercial library.

I could not find anything indicating that the repo that you are administrating is the direct successor for the codeplex site. I assumed it was a branch, but it was the branch that appeared most updated.

PR #122, opened 2016-01-26
Added support for .NET 3.5

benshoof:
SharpCompress can now be compiled for .NET 3.5 by defining symbol NET35


adamhathcock:
I probably should have removed the NET2 flags but anyway.  Do you need NET35 support?  I will target it with the xproj switch over.

Branch needs updating :)

benshoof:
I still need 3.5 but I don't expect anyone else to. Since the parts were there anyway I thought I'd just tie them together with the symbol to make it easy for me but I don't mind manually patching on my end if you'd rather get rid of them.

PR #123, opened 2016-01-26
Fix various warnings

kenkendk:
When building the solution with Xamarin Studio on OSX, it complains about a number of unused fields and variables.

Commenting these out has no ill effect. I can make an update with the fields deleted if you like.


adamhathcock:
Yes, please update with them deleted.  I really need to make a pass on this codebase with resharper as that's never been done in it's 10 years or so of existence.  Thanks!

kenkendk:
@adamhathcock done

Issue #126, opened 2016-02-09
IReader.MoveToNextEntry() not terminating


eklann:
I have encountered an issue with reading corrupt rar-files. When reading the entries in the file, the method IReader.MoveToNextEntry does not terminate, and the programs hangs. The program still consumes a lot of CPU, so it is doing something. I have tried to wait, but after 20 hours nothing has happened.

The following code can reproduce the issue:
```C#
using SharpCompress.Archive;
using SharpCompress.Reader;
using System;

namespace NonTerminatingRar
{
    internal class Program
    {
        private static void Main(string[] args)
        {
            IArchive archive = ArchiveFactory.Open(args[0]);
            using (IReader reader = archive.ExtractAllEntries())
            {
                while (reader.MoveToNextEntry())
                {
                    Console.WriteLine(reader.Entry.Key);
                }
            }
        }
    }
}
```

I am using Sharpcompress 0.11.4 from nuget.org

I also have a file that can be used to reproduce the problem, but I am not sure where to send it? It is approximately 100MB

adamhathcock:
If the RAR is corrupt, then I can't do anything about it.  SharpCompress doesn't do anything to detect corrupt entries.  It tries to extract and I guess the decompress is confused and looping.

eklann:
I understand that I cannot expect the extraction to work. But I do expect it to terminate, for example by throwing an exception.

adamhathcock:
If it's looping in the decompression, how do you propose detecting it?  It's not just a matter of throwing an exception "if something bad happens."  The problem is knowing if something bad happens.  There is nothing that detects loops like this.  

If you want, you can see if WinRAR itself detects it and perhaps suggest checksums or something is done but then again, you don't have a checksum if you can't decompress it.

eklann:
I don't know about the inner workings of rar decompression, so I don't have a solution to this issue. I am posting it since I believe it to be a problem.

WinRAR gives me a Diagnostic message when opening the file, saying "Unexpected end of archive", and after that I can extract one file from the rar. The test program I provided lists the same file as WinRAR, but hangs after that.

adamhathcock:
It might be an issue where I need to try to count the bytes read or there's a loop in header expectation.  It all depends on the type of corrupted file.  It's just not nothing I've worked on: making all corrupt scenarios work.  For better or worse, I've just been concentrating on the happy path scenarios.

eklann:
[...]

PR #127, opened 2016-02-10
Fixed bug triggered by incomplete rar file.

eklann:
Fix of this bug:
https://github.com/adamhathcock/sharpcompress/issues/126


adamhathcock:
Definitely looks wrong.  I'll check against other tests.  Thanks!

PR #143, opened 2016-05-27
New Compressione Method

alberto98fx:
Hi there! 
Thought you need a Huffman compression method...

### DIFF
--- a/SharpCompress/Compressor/Huffman.cs
+++ b/SharpCompress/Compressor/Huffman.cs
@@ -1,0 +1,226 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.IO;
+
+namespace Book
+{
+    public class Huffman16
+    {
+        private readonly double log2 = Math.Log(2);
+
+        private List<Node> HuffmanTree = new List<Node>();
+
+        internal class Node
+        {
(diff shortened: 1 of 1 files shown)


adamhathcock:
Thanks!  What applications do you see this for?  I mean, is it typically used for file compression directly?

alberto98fx:
Yeah! I think it might be useful for text compression  :+1: It is quite fast, you can test it if you want... 

This is how to use it: 
```
var freqs = Huffman16.GetFrequencies(File.ReadAllBytes(filename, true); //this is how to get frequencies it is **not** a int32, it's a long!

var huff = new Huffman16(freqs);
 
byte[] toencode = File.ReadAllBytes(filename);
           
byte[] encoded = huff.Encode(toencode); //lets encode the file! 
 
File.WriteAllBytes(filename, encoded);
```

adamhathcock:
Sorry, I'm still unsure of the PR going into SharpCompress as the Huffman stuff isn't integrated anywhere else for use.

This definitely is not a catch-all library for things like that.  The AES shortcuts you've just added don't belong.  Sorry.

weltkante:
Huffman encoding is part of [deflate](https://en.wikipedia.org/wiki/DEFLATE) - which in turn is already part of several archive formats supported by SharpCompress.

Issue #73, opened 2015-07-26
Problem Extracting 7z Archives


hodm:
There is a problem when extracting small 7z files packed with ultra compression.
Exception: Data Error StackTrace:    at SharpCompress.Compressor.LZMA.Decoder.Code(Int32 dictionarySize, OutWindow outWindow, Decoder rangeDecoder) in b:\Users\Hod\Documents\GitHub\sharpcompress\SharpCompress\Compressor\LZMA\LzmaDecoder.cs:line 357
get 7z file at: http://www.filedropper.com/mg83
inside there is a program.cs file that I used for test,
it uses the added option that I added in pull request for, but you can remove them.

adamhathcock:
I have to confess that 7z is a drop in implementation from another project.  If LZMA has changed I will unlikely be able to fix it unless another project has.  If it's a matter of reading 7z headers correctly then it's possible but I don't have the time these days to devote to this.

weltkante:
The link doesn't seem to work for me, if you can refresh it I can take a look.

(I'm the author of the managed-lzma library used in some branches of SharpCompress, but the stack trace seems to come from the lzma-sdk implementation, right? I can take a look anyways since I understand most of the workings of the 7z format and may be able to spot where it goes wrong.)

weltkante:
Surprisingly my own library also has problems reading this particular archive. Didn't expect that since it's a straight port of the official C/C++ code ... interesting, I'll take a look and try to figure out where the implementations diverge.

weltkante:
There's actually two issues with this file, one being that the file attributes are linux and not windows, thats (probably) the big red 81a48020 block going on in your screenshot. I'll need to have a look how the official codebase detects and handles linux file attributes.

Then something else is going on with the actual decoded data. I don't have figured out what is going on there, yet.

weltkante:
Updating my findings about linux file attributes: they are not part of the 7z spec and won't be written by the official 7z codebase.

Some third party implementations abused the fact that the official implementation is lazy in its error checking and started to hide linux file attributes in unused bits of the 7z metadata. Newer versions of the official 7z UI acknowledge this and try to parse the unofficial linux attributes, but it's still not part of the actual 7z archive codebase, it's only handled by the UI client.

Since it now seems to have become a kind of de-facto standard I'll update my implementation to support this unofficial extension and also prepare a pull request for SharpCompress to not treat this as a broken archive.

(Still need to look at the 2nd exception when decoding a stream.)

[...]

PR #154, opened 2016-07-20
Fix for issue #73

weltkante:
7z archives may require alternating reads from multiple substreams so it is important to seek before reading from the underlying stream. To keep performance at an acceptable level it is necessary to perform buffering because seeking on every single one-byte-read will destroy performance.


adamhathcock:
Fine with me.  I really hate 7zip.

PR #168, opened 2016-09-09
Adds support for Apple Data Compression.

claunia:
Adds support for Apple Data Compression algorithm as a stream and byte[].


src/SharpCompress/Compressor/ADC/ADCBase.cs
```diff
@@ -0,0 +1,210 @@
...
+                        {
+                            full = true;
+                            break;
+                        }
+                        input.Read(buffer, outPosition, chunkSize);
+                        outPosition += chunkSize;
+                        position += chunkSize + 1;
+                        break;
+                    case TwoByte:
+                        tempMs = new MemoryStream();
```
adamhathcock, line 137: Looks like these  temp MemoryStreams aren't required.  Just write directly to the output or a simple array of two or three bytes?

src/SharpCompress/Compressor/ADC/ADCBase.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        /// <summary>
+        /// Decompresses a stream that's compressed with ADC
+        /// </summary>
+        /// <param name="input">Stream containing compressed data</param>
+        /// <param name="output">Buffer to hold decompressed data</param>
+        /// <param name="bufferSize">Max size for decompressed data</param>
+        /// <returns>How many bytes are stored on <paramref name="output"/></returns>
+        public static int Decompress(Stream input, out byte[] output, int bufferSize = 262144)
```
adamhathcock, line 98: Rather than this method signature, couldn't this be the same as in `Read(byte[], int, int)`?

Why buffer at all?  Let the Stream contract control the buffer size.  There's a `BufferedStream` class that handles buffering if performance is improved by having a larger buffer size.

adamhathcock:
Overall looks good to me.  Just some questions to try to reduce allocations and copying if possible.

One thing I want to do overall to the library is to a buffer pool from .NET Core's System.Buffers package.  I never have time in between work and kids now :/

src/SharpCompress/Compressor/ADC/ADCBase.cs
claunia: As I said, these algorithms were used for disk images, and the exact buffer size is known previous to decompressing. Also this particular algorithm rewinds, and there is no warranty that a stream can be rewind.

src/SharpCompress/Compressor/ADC/ADCBase.cs
claunia: That is possible.

src/SharpCompress/Compressor/ADC/ADCBase.cs
adamhathcock: You want to keep things streamable to use less memory.  There's no reason to have a very large buffer in memory from what I see.

I could be misunderstanding though.

claunia: I have tried several ways to stream it, even less buffer sizes, but didn't get it to decompress correctly, or it directly crashed. I don't consider 256KiB to be a big buffer, even less a "very large", and don't worry, last time this algorithm was used in the wild was when 128MiB or RAM were standard.

If you can find a way to do it more streamable I'm open to suggestions, but I tried several ones and all failed miserably, and all other implementations use a buffer also, so I'm out of ideas.

adamhathcock: It's fine really. I'm just offering possible suggestion and probably being overly pandatic.

claunia: No problem.

Issue #166, opened 2016-09-02
Can't extract big tbz file(408 MB)


agoretsky:
Hello! I am trying to extract this file:
https://drive.google.com/open?id=0B_hcLdxD3U8Td2dXRGhSaHl0cDg
In destination folder I've got only 879 KB of result and it is not growing, but program is still running.
Looks like program  is stuck or all archive is still extracting in memory (but I can't see process with growing memory)
My code:
```
        private void ExtractArchiveFileToTempFolder(string path, string feedDate)
        {
            var destination = Path.Combine(this.tempFolderPath, feedDate);
            if (!Directory.Exists(destination))
            {
                Directory.CreateDirectory(destination);
            }

            using (Stream stream = File.OpenRead(path))
            {
                var reader = ReaderFactory.Open(stream);
                while (reader.MoveToNextEntry())
                {
                    if (!reader.Entry.IsDirectory)
                    {
                        Console.WriteLine(reader.Entry.Key);
                        reader.WriteEntryToDirectory(destination);
                    }
                }
            }
        }
```

adamhathcock:
Got this file and it doesn't seem winrar can open this either.

SharpCompress doesn't detect errors.

erikcturner:
It's true that WinRAR can't open the file.  But 7zip opens the file just fine.  I was able to extract a file called collection.tar.  And then I was able to extract a file "itunes20160831\collection" that was 2,126,385,940 bytes long.

I suspect that this is the same error that I was having (see #165 submitted on Aug 28).  The issue for me was that the the ".tar.gz" file was truncated during transfer and then caused the SharpCompress decoder to hang (throwing an exception would definitely have been better).

adamhathcock:
This looks like my BZip2 implementation only extracts 879k.  WinRAR doesn't like it either.  I guess 7Zip does.

agoretsky:
I am using WinRar 5.40 beta, right click on archive then "Extract files" and everything works fine

adamhathcock:
Actually, I see that too now.  Just opening it in the UI doesn't look right.

I'll do some more investigation tomorrow

adamhathcock:
This is weird.  All implementations of Bzip2 in C# I've tried will only extract 879k of the file.  It just stops after that.

plusulica:
Hi Adam.
Do you have any news about your last comment? Also me I got the same weird behaviour, everything that I try to decompress is truncated to 879 kb. Thanks in advance.

[...]

PR #179, opened 2016-09-28
Allow empty tar header to be read to know there are no more tar heade…

adamhathcock:
…rs to read


adamhathcock:
This was found while debugging https://github.com/adamhathcock/sharpcompress/issues/166

PR #183, opened 2016-09-30
Adds support for Apple DART's RLE, corrects comments and copyright messages.

claunia:
This adds support for Apple DART's RLE algorithm.

I've put it in its separate folder because DART implements another algorithm, a variation of LZH differently enough that no known implementation decodes it correctly.

Or maybe we should move DART's RLE/LZH, ADC, and in the future, KenCode, to a folder called "Apple"???

I also corrected the comments that contained incorrect copyright messages and filenames, and the Formats documentation.


src/SharpCompress/Compressors/DART/RLEStream.cs
```diff
@@ -0,0 +1,235 @@
...
+                throw new ArgumentOutOfRangeException(nameof(offset));
+            }
+            if ((offset + count) > buffer.GetLength(0))
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
+            }
+
+            int copied = 0;
+
+        fillBuffer:
```
adamhathcock, line 163: Anyway you can do this without the `goto` and label?  Just bad C# style.

adamhathcock:
Looks mostly good.  I think you need to resolve this against the latest master for some reason.  Hopefully then the PR build will work.

src/SharpCompress/Compressors/DART/RLEStream.cs
claunia: Mayyyyyyybe, I need to test.

And while most C# people tend to ignore gotos, they're used extensively in .NET frameworks and the IL optimizes them easy.

But if you prefer no goto I'll try to remove it, don't worry.

claunia:
The only conflict is because of FORMATS.md *sigh*.
I don't know if github allows you do something about it, if not well I'll rebase with latest master.

adamhathcock:
You can't resolve conflicts in github itself.  You can just pull master into your branch again (resolve the conflict) and it should be fine.

src/SharpCompress/Compressors/DART/RLEStream.cs
adamhathcock: It's not a big deal.  I may refactor later if you don't want to :)

claunia: Don't worry I'll do it myself this week and with some luck I add compression.

Issue #197, opened 2016-11-18
Pkware encryption for Zip files didn't allow for multiple reads of an entry


alexmgerasimov:
Hello!

I tried to use this library and faced some issues. Basically repeated reading of some zip archive entries leads to crashes. It seems that this is related to encrypted zips somehow. I'm using the following test code to reproduce the crashes:

```
	class Program
	{
		static void Main(string[] args)
		{
			try
			{
				using (var fileStream = File.Open(@"c:\Archive1.zip", FileMode.Open))
					UnpackMultipleTimes(fileStream, "12345678");
			}
			catch (Exception e)
			{				
			}
		}

		private static void UnpackMultipleTimes(Stream stream, string password)
		{
			using (var archive = ArchiveFactory.Open(stream, new ReaderOptions {Password = password}))
			{
				var entries = archive.Entries.Where(entry => !entry.IsDirectory);

				foreach (var entry in entries)
				{
					for (int i = 0; i < 100; i++)
					{
						using (var memoryStream = new MemoryStream())
						using (var entryStream = entry.OpenEntryStream())
						{
							entryStream.CopyTo(memoryStream);
						}
					}
				}
			}
		}
	}
```

Exception from [Archive1.zip](https://github.com/adamhathcock/sharpcompress/files/600724/Archive1.zip):

> SharpCompress.Compressors.Deflate.ZlibException was unhandled
>   HResult=-2146233088
>   Message=Bad state (invalid block type)
>   Source=SharpCompress
>   StackTrace:
>        в SharpCompress.Compressors.Deflate.InflateManager.Inflate(FlushType flush)
>        в SharpCompress.Compressors.Deflate.ZlibBaseStream.Read(Byte[] buffer, Int32 offset, Int32 count)
>        в System.IO.Stream.InternalCopyTo(Stream destination, Int32 bufferSize)
>        в SevenZipTest.Program.UnpackMultipleTimes(Stream stream, String password)
>        в SevenZipTest.Program.Main(String[] args)
>        в System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
>        в System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
>        в Microsoft.VisualStudio.HostingProcess.HostProc.RunUsersAssembly()
>        в System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[...]

PR #200, opened 2016-11-24
Open branch, to fix multiple crashes on repeated zip archives reading…

vladimdemidov:
…. Added fix.


adamhathcock:
Thanks!

I think I see what the changes are but can you make the spacing consistent?

Also, other minor things.

SharpCompress.sln
```diff
@@ -1,44 +1,44 @@
-﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
```
adamhathcock, line 2: Line endings changed?

src/SharpCompress/Common/Zip/ZipFilePart.cs
```diff
@@ -124,40 +124,46 @@ protected Stream CreateDecompressionStream(Stream stream)
...
+						return new PkwareTraditionalCryptoStream(plainStream, Header.ComposeEncryptionData(plainStream), CryptoMode.Decrypt);
+
+					case ZipCompressionMethod.WinzipAes:
 #if !NO_FILE
-            if (Header.WinzipAesEncryptionData != null)
-            {
-                //only read 10 less because the last ten are auth bytes
-                return new WinzipAesCryptoStream(plainStream, Header.WinzipAesEncryptionData, Header.CompressedSize - 10);
-            }
+						return Header.WinzipAesEncryptionData != null ? new WinzipAesCryptoStream(plainStream, Header.WinzipAesEncryptionData, Header.CompressedSize - 10) : plainStream;
```
adamhathcock: Why do this?

I hate the tertiary operator.  It's not readable on a single line here.

src/SharpCompress/Common/Zip/Headers/ZipFileEntry.cs
```diff
@@ -65,9 +65,28 @@ protected byte[] EncodeString(string str)
 
         internal List<ExtraData> Extra { get; set; }
 
-        internal PkwareTraditionalEncryptionData PkwareTraditionalEncryptionData { get; set; }
+        //internal PkwareTraditionalEncryptionData PkwareTraditionalEncryptionData { get; set; }
```
adamhathcock: Don't leave commented out code, please.

adamhathcock:
Fix for https://github.com/adamhathcock/sharpcompress/issues/197 I think.

SharpCompress.sln
vladimdemidov: Can you please clarify, what newline character do you use?

adamhathcock: Maybe it's still left over from before I tried to normalize endings but I have gitattributes set to auto.

Question is: why is this file changing at all?

vladimdemidov: The solution file was regenerated by Visual Studio. I don't see any real changes.

src/SharpCompress/Common/Zip/ZipFilePart.cs
vladimdemidov: Fixed.

src/SharpCompress/Common/Zip/Headers/ZipFileEntry.cs
vladimdemidov: Fixed.

SharpCompress.sln
adamhathcock: So revert it then please.  Whitespace changes make a lot of noise for reviewing.

adamhathcock:
Please revert all the whitespace changes.  You've spaced in way too far on some files for no reason.

src/SharpCompress/Common/Zip/ZipFilePart.cs
```diff
@@ -126,16 +126,11 @@ protected Stream CreateDecompressionStream(Stream stream)
...
         protected Stream GetCryptoStream(Stream plainStream)
         {
-            if ((Header.CompressedSize == 0)
-#if !NO_CRYPTO
- && ((Header.PkwareTraditionalEncryptionData != null)
-                    || (Header.WinzipAesEncryptionData != null)))
-#else
-                && (Header.PkwareTraditionalEncryptionData != null))
-#endif
+            if ((Header.CompressedSize == 0) && !string.IsNullOrEmpty(Header.Password))
```
adamhathcock: This doesn't make sense.  The lack of a password should be it's own `if` statement.  Not `AND`ed with this check.

src/SharpCompress/Common/Zip/ZipHeaderFactory.cs
```diff
@@ -165,20 +162,16 @@ private void LoadHeader(ZipFileEntry entryHeader, Stream stream)
             switch (mode)
             {
                 case StreamingMode.Seekable:
-                {
```
adamhathcock, line 168: don't remove braces please.  They're there for a reason.

SharpCompress.sln
vladimdemidov: Sorry i can't revert this file because my Git didn't see any changes.

src/SharpCompress/Common/Zip/ZipFilePart.cs
```diff
@@ -126,16 +126,11 @@ protected Stream CreateDecompressionStream(Stream stream)
...
         protected Stream GetCryptoStream(Stream plainStream)
         {
-            if ((Header.CompressedSize == 0)
-#if !NO_CRYPTO
- && ((Header.PkwareTraditionalEncryptionData != null)
-                    || (Header.WinzipAesEncryptionData != null)))
-#else
-                && (Header.PkwareTraditionalEncryptionData != null))
-#endif
+            if (Header.CompressedSize == 0)
```
adamhathcock: I guess this is my bad while this is busting things.

I can't think through the logic at the moment but the `null` checks above where used to see if encryption was being attempted.

I guess that's what you were doing with the password being specified.  Sorry, I guess that needs to be put back as I misread.

vladimdemidov: I think that define that file is encrypted can be checked by using Header.Flags. I think it easiest way, because in previous code variant, encryptor in HederFactory is created based on this flag.

PR #205, opened 2017-01-24
Add zip64 support for ZipArchive extraction

markryd:
I've taken a first cut at some limited zip64 support. It is extract only and only on the Archive API at this point.

Note: I'm casting 8 byte values to `long` instead of supporting the full `ulong` as it simplifies the implementation quite a bit. Things like `Reader.Seek` take a `long` directly, it's easy enough to cover that with multiple reads but I ended up backing away as it made the change much larger for minimal benefit (9 Exabyte archive size limitation instead of 18 etc). Also CLS compliance means we can't expose `ulong`. More than happy to revisit that if required.


adamhathcock:
Looks good to me, thanks!

Does CLS compliance matter a lot?  I guess if there is a demand for 9+ exabyte files then that can be another fix :)

adamhathcock:
Published https://www.nuget.org/packages/sharpcompress/0.15.0

PR #206, opened 2017-01-25
Zip64 extending information and ZipReader

markryd:
This PR adds support for reading the zip64 extended information field, which wasn't required for my previous test files, but probably is required in general for zip64 archives (the tripping point in my testing was new files starting beyond the 4Gb point of my archive).

Adding this support also instantly fixed the trouble I was having with ZipReader support, so that has been included.


adamhathcock:
Excellent.  Thanks!

PR #210, opened 2017-03-09
Fix invalid headers

kenkendk:
This fixes a bug introduced with 6be6ef0b5c3ed6a4d40d00c8fb133518e75e4a6f.

Bad news is that all zip files written with 0.15.1 are semi-broken.
That is, their filenames are garbled, but content is recoverable.


kenkendk:
As far as I can see the code here is not used, so my panic was unjustified.

For some reason, the ZipHeader has an abstract `Write` method, and all the subclasses implement it, but none of them call it.
This would only cause problems if someone were to manually construct the items and call the write methods.

PR #217, opened 2017-04-01
Added Support For .Net Framework 2.0

fox564:
Added Support For .Net Framework 2.0 because lets face it, its still being used.

Todo:
-Convert to vb.net

Why Do i need vb.net while i can just add it as a reference?
well some developers like me do not like using other dll files to avoid file not found or using it as embedded reference and load it in runtime.


adamhathcock:
2.0 is no longer supported by Microsoft.  3.5 SP1 is though.

As for VB.NET support, I don't know what you're saying is needed and can't see in this PR.

PR #220, opened 2017-04-04
verify RAR crc on header and file data

coderb:
verify rar file crc checksums:
- header records
- decompressed file payload

tested and works with:
- typical rar files
- rar files with encrypted headers
- multipart rar files

### DIFF
--- a/src/SharpCompress/Common/Rar/Headers/RarHeader.cs
+++ b/src/SharpCompress/Common/Rar/Headers/RarHeader.cs
@@ -1,4 +1,5 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using SharpCompress.IO;
 
 namespace SharpCompress.Common.Rar.Headers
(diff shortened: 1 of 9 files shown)


coderb:
Hi Adam,

I haven't used cake much and am still on VS 2015.  Can you help me figure out which test file is failing on the AppVeyor build?  I looks like something fails and causes a cascade of errors.

Thanks,
Brien

adamhathcock:
If you look at the AppVeyor output, it will show the failing unit test names: https://ci.appveyor.com/project/adamhathcock/sharpcompress/build/0.15.120

I haven't converted this to VS2017 csproj from VS2015 project JSON so you should be fine.  I think the xunit tests should still work with VS2015.

You can run the tests just by running the powershell script `.\build.ps1`

coderb:
sorry to seem clueless, but i've only ever used *.csproj files and my vs2015 won't load the test project which is xproj/json.

adamhathcock:
Download the VS2015 tooling from here: https://www.microsoft.com/net/download/core

coderb:
i found the issue which turned up a (minor?) bug in the existing code. 
The bug is that RarCryptoBinaryReader buffers bytes from the underlying stream and the CurrentReadByteCount is being updated to match the buffered consumption rather than the end use.

I will repush a commit when i get the code cleaned up.  My fix requires CurrentReadByteCount and Mark to become virtual.  Are you ok with that or would you like to attempt to refactor differently?

coderb:
hurray, appveyor is happy!  adam- i'm done for now.  please review and supply feedback if there are any changes you want made prior to merging.  cheers, b!

(btw, i use this library heavily and am very appreciative of it's existence!  thanks for the hard work!)

PR #225, opened 2017-04-20
Add new feature to allow injection of an action into the extraction process to log and show progress.

gardebring:
Add new feature to allow injection of an action into the extraction process. This allows for showing or logging progress of the extraction process, especially useful for large files that might take a long time to extract.

### DIFF
--- a/src/SharpCompress/Archives/IArchiveEntryExtensions.cs
+++ b/src/SharpCompress/Archives/IArchiveEntryExtensions.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.IO;
+﻿using System.IO;
 using SharpCompress.Common;
 using SharpCompress.IO;
 using SharpCompress.Readers;
@@ -8,7 +7,7 @@ namespace SharpCompress.Archives
 {
     public static class IArchiveEntryExtensions
     {
-        public static void WriteTo(this IArchiveEntry archiveEntry, Stream streamToWriteTo, Action<long, int> partTransferredAction = null)
+        public static void WriteTo(this IArchiveEntry archiveEntry, Stream streamToWriteTo)
         {
             if (archiveEntry.Archive.Type == ArchiveType.Rar && archiveEntry.Archive.IsSolid)
             {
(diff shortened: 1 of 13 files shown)


adamhathcock:
There already events on Readers and Archives like `FireFilePartExtractionBegin` they may not cover what you're doing but there shouldn't be two mechanisms to listen to progress.

Can you use those events or do they not accomplish what you need?  You can see the listener interface passed around where you've modified the code.

gardebring:
Hi Adam and thanks for your feedback. An event is, as you stated, indeed probably a better idea. It's a large codebase and I am not completely familiar with it yet, so I felt safer doing it the way I did, but I will have another look and see if I can do this as an event instead.

gardebring:
Alright, so a bit of an update here.
I can access the StreamListener from within AbstractReader, and it's easy to hookup a new event here.
However, once we move focus onto TransferTo in Utility we do loose that context and we cannot introduce the StreamListener into TransferTo since that method is used by other callers as well.

So my suggestion is: 
I add a new EventHandler to IReader
I add a new Fire method to I IReaderExtractionListener to invoke this new event.
I pass the streamListener created in the WriteEntryTo method to the Write Method.
However, as a final step, I will not pass the streamListener to the TransferTo method (since this would introduce a dependency I think should be avoided), but instead I will here pass an optional Action, basically keeping the TransferTo method as my initial suggestion but moving all logic surrounding it so instead of passing down an action all the day we fetch it from the new event instead.

adamhathcock:
I have questions but I guess I want to see what you try.

I'd like to have the `IStreamListener` interface be holistic in that it's the only thing used for events.  How that looks can be up to you.

gardebring:
Hi again Adam. Please check out this commit and see if its more in line with your thoughts:
https://github.com/gardebring/sharpcompress/commit/683d2714d0b058b570ba4cbcda3e97c2cd4968e4

This allows me to assign an event to the reader as such:
`reader.EntryExtractionProgress += Reader_EntryExtractionProgress;`

And then use it as such:
`        
private int lastPercentage = -1;
//---//

private void Reader_EntryExtractionProgress(object sender, ReaderExtractionEventArgs<IEntry> e)
        {
            var progress = (SharpCompress.Readers.ReaderProgress)e.ParamList[0];
            if (progress.PercentageRead % 10 == 0 && lastPercentage != progress.PercentageRead)
            {
                // Do something for every 10 percent of the extraction
            }
        }
`

PR #226, opened 2017-04-25
Add new event handler to allow tracking of progress of extraction progress for individual entry

gardebring:
Add new event handler to allow tracking of progress of extraction progress for individual entry. This allows for showing or logging progress of the extraction process, especially useful for large files that might take a long time to extract.

This allows us to assign an event to the reader as such:
`reader.EntryExtractionProgress += Reader_EntryExtractionProgress;`

And then for example use it as such:
`        
private int lastPercentage = -1;
//---//

private void Reader_EntryExtractionProgress(object sender, ReaderExtractionEventArgs<IEntry> e)
        {
            var progress = (SharpCompress.Readers.ReaderProgress)e.ParamList[0];
            if (progress.PercentageRead % 10 == 0 && lastPercentage != progress.PercentageRead)
            {
                // Do something for every 10 percent of the extraction
            }
        }
`

### DIFF
--- a/src/SharpCompress/Common/ReaderExtractionEventArgs.cs
+++ b/src/SharpCompress/Common/ReaderExtractionEventArgs.cs
@@ -1,16 +1,17 @@
 ﻿using System;
+using SharpCompress.Readers;
 
 namespace SharpCompress.Common
 {
     public class ReaderExtractionEventArgs<T> : EventArgs
     {
-        internal ReaderExtractionEventArgs(T entry, params object[] paramList)
+        internal ReaderExtractionEventArgs(T entry, ReaderProgress readerProgress = null)
         {
             Item = entry;
-            ParamList = paramList;
+            ReaderProgress = readerProgress;
         }
 
         public T Item { get; private set; }
-        public object[] ParamList { get; private set; }
+        public ReaderProgress ReaderProgress { get; private set; }
     }
 }
--- a/src/SharpCompress/Common/ReaderExtractionEventArgs.cs
+++ b/src/SharpCompress/Common/ReaderExtractionEventArgs.cs
@@ -4,11 +4,13 @@ namespace SharpCompress.Common
 {
     public class ReaderExtractionEventArgs<T> : EventArgs
     {
-        internal ReaderExtractionEventArgs(T entry)
+        internal ReaderExtractionEventArgs(T entry, params object[] paramList)
         {
             Item = entry;
+            ParamList = paramList;
         }
 
         public T Item { get; private set; }
+        public object[] ParamList { get; private set; }
     }
 }
--- a/src/SharpCompress/Readers/AbstractReader.cs
+++ b/src/SharpCompress/Readers/AbstractReader.cs
@@ -191,7 +191,7 @@ namespace SharpCompress.Readers
         {
             using (Stream s = OpenEntryStream())
             {
-                s.TransferTo(writeStream, (sizeTransferred, iterations) => streamListener.FireEntryExtractionProgress(Entry, sizeTransferred, iterations));
+                s.TransferTo(writeStream, Entry, streamListener);
             }
         }
 
--- a/src/SharpCompress/Readers/AbstractReader.cs
+++ b/src/SharpCompress/Readers/AbstractReader.cs
@@ -167,7 +167,7 @@ namespace SharpCompress.Readers
             }
         }
 
-        public void WriteEntryTo(Stream writableStream)
+        public void WriteEntryTo(Stream writableStream, Action<long, int> partTransferredAction = null)
         {
             if (wroteCurrentEntry)
             {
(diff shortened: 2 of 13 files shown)


adamhathcock:
Why did you close the other PR to recreate it?  My review comments are still the same: https://github.com/adamhathcock/sharpcompress/pull/225

gardebring:
Hi. The code is completely rewritten so I thought it made sense to close the existing one and create a new instead. Sorry if it was confusing, was not my intention. The review comments in #225 is not really valid anymore since the code is now rewritten using an event handler instead.

adamhathcock:
It's the same code on the same branch.  Did you mean to make this pull request from a different branch other than `gardebring:master` ?

gardebring:
Sorry but you lost me now.
The old code used an action as a parameter passed along. The new code use an event handler, so it is not the same code.
However, the PR do contain the old commits that were reverted. 
So, first commit is old code, second is revert, third is the new code.
I can create a completely new PR with only the final change instead if you want. It will not have a functional different, but the  commit history will look nicer.

adamhathcock:
So the code is correct as you only have your master branch.  Pull Requests work per branch so it was updated when you pushed the latest changes.

My review from the other Pull Request is the same as this Pull Request as it's the same branch with the same code.

Ill just move over my comments.

src/SharpCompress/Common/ReaderExtractionEventArgs.cs
```diff
@@ -4,11 +4,13 @@ namespace SharpCompress.Common
...
     {
-        internal ReaderExtractionEventArgs(T entry)
+        internal ReaderExtractionEventArgs(T entry, params object[] paramList)
         {
             Item = entry;
+            ParamList = paramList;
         }
 
         public T Item { get; private set; }
+        public object[] ParamList { get; private set; }
```
adamhathcock: Make this strongly typed. I don't see a need for this as is.

src/SharpCompress/Readers/AbstractReader.cs
```diff
@@ -181,16 +182,16 @@ public void WriteEntryTo(Stream writableStream)
...
             wroteCurrentEntry = true;
         }
 
-        internal void Write(Stream writeStream)
+        internal void Write(Stream writeStream, IReaderExtractionListener streamListener)
         {
             using (Stream s = OpenEntryStream())
             {
-                s.TransferTo(writeStream);
+                s.TransferTo(writeStream, (sizeTransferred, iterations) => streamListener.FireEntryExtractionProgress(Entry, sizeTransferred, iterations));
```
adamhathcock: I guess this is okay. I wouldn't have a problem just duplicating the transfer code to have listening and not listening.

Kind of doesn't make sense to have a lambda just to call the listener when the listener can just be passed in.

src/SharpCompress/Readers/IReader.cs
```diff
@@ -8,6 +8,7 @@ public interface IReader : IDisposable
     {
         event EventHandler<ReaderExtractionEventArgs<IEntry>> EntryExtractionBegin;
```
adamhathcock: I would dump the Begin and End and only have Progress. They can check iterations or something to see if it's started.

src/SharpCompress/Readers/ReaderProgress.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace SharpCompress.Readers
+{
+    public class ReaderProgress
+    {
+        private readonly IEntry _entry;
+        public long BytesTransferred { get; private set; }
+        public int Iterations { get; private set; }
+
+        public int PercentageRead => (int)Math.Round(PercentageReadExact);
+        public double PercentageReadExact => (float)BytesTransferred / _entry.Size * 100;
```
adamhathcock, line 15: I like the way this was done

src/SharpCompress/Utility.cs
```diff
@@ -229,15 +229,18 @@ public static DateTime DosDateToDateTime(Int32 iTime)
             return DosDateToDateTime((UInt32)iTime);
         }
 
-        public static long TransferTo(this Stream source, Stream destination)
+        public static long TransferTo(this Stream source, Stream destination, Action<long, int> action = null)
```
adamhathcock: Again, I would just pass IStreamListener as an optional value like this.

gardebring:
Hi again and thanks for feedback. For some strange reason I could not see your code comments previously, which probably explains the confusion. I will look into them and get back to you.

src/SharpCompress/Readers/IReader.cs
gardebring: I sort of agree, but this would break current implementations using those event handlers. Would this be ok?

src/SharpCompress/Utility.cs
gardebring: As I stated, my thought process was to not introduce a new dependency into utility, but if you are fine with this I will change to that behavior.

src/SharpCompress/Common/ReaderExtractionEventArgs.cs
gardebring: The idea here was to generalize what could be passed, but am I correct in assuming that you want ReaderProgress here instead of object[] ?

adamhathcock: Yeah.  We'll worry about other options later.  No need to over genericize at the moment I think.

src/SharpCompress/Readers/IReader.cs
adamhathcock: Yeah, I break things all the time :)

I keep this library "beta" just because I haven't given things enough thought nor have the time to worry about API breaking on updates.  Users should expect to at least recompile to find API breakages.

src/SharpCompress/Utility.cs
adamhathcock: This Utility won't be reused and it's just for the library.  I'm fine with internal dependencies.

gardebring:
Alright, so I believe the requested changes now has been implemented. Let me know if there is anything else :)

PR #266, opened 2017-07-05
Force encoding

eklann:
Added property ForceEncoding to OptionsBase, purpose is to be able to force reading from or writing to an archive to use the specified encoding.

The problem I am solving with this is the ability to unpack archives that has a non-standard encoding of the filenames. This is common for zip files.

The ArchiveEncoding.Default does not solve the problem, because of two reasons.
1. It is global and can not be set per archive, and when multiple threads are used to unpack archives this is just not possible to use.
2. If a zip file has the UTF-8-flag set AchiveEncoding.Default is not used, and I need to be able to force the use of a specific encoding regardless of any flags.


adamhathcock:
I like this.  Thanks!

I never liked having the static encoding but I just didn't put it on options for some reason.  I'll probably follow this up with removing the static altogether.

Try not to mix in so many whitespace changes with real changes next time.

eklann:
Sorry about the intertwined whitespace, will take care next time. Thanks for the fast response! 

Do you know when I can expect a release of this on nuget.org?

adamhathcock:
I'm on holiday for a month currently so I don't know when I'll get a chance to finish up and release something.

Issue #269, opened 2017-07-07
If Zip file has normal file header AND a post-descriptor header AND the file is attempted to be skipped by a ZipReader, then the data is attempted to be skipped twice.


KvanTTT:
I have the following [BadStateArchive.zip](https://github.com/adamhathcock/sharpcompress/files/1130352/BadStateArchive.zip).

I want to ignore files that start with the dot: `.DS_Store`. But if I just ignore such file I get the exception 
`Bad state (invalid stored block lengths)` in `reader.MoveToNextEntry()` method. I use the following code for reading:

```CSharp
using (Stream stream = File.OpenRead(sourceArchiveFileName))
{
    using (IReader reader = ReaderFactory.Open(stream))
    {  
        while (reader.MoveToNextEntry())
        {
           ...
        }
    }
}
```

If such files writing to `DummyStream` exception does not occur:

```CSharp
using (var dummyStream = new DummyStream())
{
    reader.WriteEntryTo(dummyStream);
}
```

where `DummyStream` is

```CSharp
public class DummyStream : Stream
{
    public override bool CanRead => true;

    public override bool CanSeek => true;

    public override bool CanWrite => true;

    public override long Length => 0;

    public override long Position { get; set; } = 0;

    public override void Flush()
    {
    }

    public override int Read(byte[] buffer, int offset, int count)
    {
        return 0;
    }

    public override long Seek(long offset, SeekOrigin origin)
    {
        return 0;
    }

    public override void SetLength(long value)
    {
    }

    public override void Write(byte[] buffer, int offset, int count)
    {
    }
}

```

adamhathcock:
It's probably because you're returning 0 on Read which means the stream is empty.  I don't think decompression likes trying to operation on a zero byte stream.

KvanTTT:
But I should not use `WriteEntryTo` method for getting rid of exception.

adamhathcock:
I don’t know offhand but I may just have a try/catch that’s hiding that exception in that extension.

KvanTTT:
I tried to wrap a `reader.MoveToNextEntry()` to try/catch block. But in this case, not all files being processed. Although they should.

adamhathcock:
Sorry.  I finally understand what you're trying to: just skip a .DS_Store file but it isn't working.

The dummy stream stuff you mentioned doesn't have anything to do with it.

PR #272, opened 2017-07-08
Don't skip ZipReader data twice.

adamhathcock:
This could occur when a zip file has a normal file header AND a post descriptor.

Fixes https://github.com/adamhathcock/sharpcompress/issues/269


KvanTTT:
Could you also add a test for BadStateArchive.zip from #269 issue?

adamhathcock:
Done

PR #278, opened 2017-07-15
Optional string decoder delegate: fix #277

### LINKED ISSUE #277: Add Entry.GetKeyBytes() method
Related to #163, #261 and #266.

Since the string encoding in Zip archive is such a great pain, is it possible to provide a `byte[]` which contains raw bytes of file name without decoding, so the user can employ other libs (e.g. [Chardet](https://github.com/errepi/ude)) to process it later?

If you think it's okay, I am happy to do it and then open a PR :-)

xupefei:
A simple usage with [SimpleHelpers.FileEncoding](https://www.nuget.org/packages/SimpleHelpers.FileEncoding/) package (which is a Chardet helper):
```
var f = ArchiveFactory.Open(@"X:\some-really-old.zip", new ReaderOptions
{
    ArchiveEncoding = new ArchiveEncoding
    {
        CustomDecoder = (bytes, start, count) =>
            FileEncoding.DetectFileEncoding(bytes, start, count, Encoding.UTF8)
                .GetString(bytes, start, count)
    }
});
foreach (var entry in f.Entries)
{
    Debug.WriteLine(entry.Key);
}
```

### DIFF
--- a/src/SharpCompress/Common/ArchiveEncoding.cs
+++ b/src/SharpCompress/Common/ArchiveEncoding.cs
@@ -14,11 +14,17 @@ namespace SharpCompress.Common
         /// ArchiveEncoding used by encryption schemes which don't comply with RFC 2898.
         /// </summary>
         public Encoding Password { get; set; }
-        
+
         /// <summary>
-        /// Set this encoding when you want to force it for all encoding/decoding operations.
+        /// Set this encoding when you want to force it for all encoding operations.
         /// </summary>
         public Encoding Forced { get; set; }
+
+        /// <summary>
+        /// Set this when you want to use a custom method for all decoding operations.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


adamhathcock:
Thanks this is good.

I wonder if I should take the step of updating, distributing and depending upon the Chardet implementation directly.

xupefei:
Better not: Chardet is not always correct (*1, *2) and may not able to run on .NET Standard 1.X.
However I think it is good to *recommend* user to use Chardet, rather than embed it.

*1: Chardet will return false answer for short string e.g. `abcdefg中文`. It reports CP1252 but in fact it is CP 936.
*2: Chardet does not support all languages: https://github.com/errepi/ude

adamhathcock:
I may make it a dependency after updating it.  Separate to this project.

But yes, auto-detection may not always be good. 

Thanks!

PR #287, opened 2017-08-16
Fixed bug: Passing default ReaderOptions when creating ZipReader for solid extraction

twirpx:
This bug causes entry keys in wrong encoding when using ExtractAllEntries/MoveToNextEntry if non-UTF8 encoding specified at ZipArchive.Open


adamhathcock:
Thanks!

Issue #298, opened 2017-09-15
ZlibStream fails to check checksum correctly


ied206:
ZlibStreams occasionally fails to decompress correct zlib files.

.zz files in [TestSample.zip](https://github.com/adamhathcock/sharpcompress/files/1307402/TestSample.zip) are the zlib streams which causes ZlibStream to malfunction.
When ZlibStream tries to decompress those files, it throws ZlibException with 'invalid data check' message.

However, those files can be successfully extracted in [pigz](https://zlib.net/pigz/) with --zlib flag, so the problem resides in SharpCompress.

To prove this, I converted zlib streams into pure deflate streams by removing first 2 byte (zlib magic number) and last 4 byte (Adler32 checksum), and tried to decompress with DeflateStream. In this case, file was successfully decompressed, so I suspect SharpCompress' Adler32 checksum checking routine is faulty.

I wrote some [Tests](https://github.com/ied206/SharpCompressBug) to help debugging, please have a look.

### Tested Environment
OS : Windows 10 x64 v1703
IDE : Visual Studio 2017
DotNet : .Net Framework 4.6.2, .Net Framework 4.7
Package : SharpCompress 0.18.1

ied206:
More addtional information, I checked the internal adler32 checksum in InflateBlocks.Process().

### ex1_zlib_sharpcompress.jpg.zz
- computedCheck : 0xd2367044
- expectedCheck : 0xd77c7044

### ex2_zlib_sharpcompress.jpg.zz
- computedCheck : 0x99d5edad
- expectedCheck : 0x9b97edad

So only the high two bytes are miscalculated.

adamhathcock:
Fixed by https://github.com/adamhathcock/sharpcompress/pull/301

Issue #296, opened 2017-09-12
Block header corrupt on correct XZ file


claunia:
`Unhandled Exception:
System.IO.InvalidDataException: Block header corrupt
  at SharpCompress.Compressors.Xz.XZBlock.CacheHeader () [0x00060] in <57fade50bf4b4b0aa2c9652265c0bd24>:0 
  at SharpCompress.Compressors.Xz.XZBlock.LoadHeader () [0x00006] in <57fade50bf4b4b0aa2c9652265c0bd24>:0 
  at SharpCompress.Compressors.Xz.XZBlock.Read (System.Byte[] buffer, System.Int32 offset, System.Int32 count) [0x0000a] in <57fade50bf4b4b0aa2c9652265c0bd24>:0 
  at SharpCompress.Compressors.Xz.XZStream.ReadBlocks (System.Byte[] buffer, System.Int32 offset, System.Int32 count) [0x0001f] in <57fade50bf4b4b0aa2c9652265c0bd24>:0 
  at SharpCompress.Compressors.Xz.XZStream.Read (System.Byte[] buffer, System.Int32 offset, System.Int32 count) [0x0001a] in <57fade50bf4b4b0aa2c9652265c0bd24>:0 `

File is https://download.freenas.org/9.2.1/RELEASE/x86/FreeNAS-9.2.1-RELEASE-x86.vmdk.xz

claunia:
Thanks!

Issue #289, opened 2017-08-18
'Block header corrupt' when decompressing xz file


OtherCrashOverride:
I receive a System.IO.InvalidDataException with the above message (Block header corrupt) when trying to decompress a .xz file.  It occurs at SharpCompress.Compressors.Xz.XZBlock.CacheHeader().

The xz file decompresses without issue using 7zip.  The file tested can be found here:
https://github.com/ayufan-rock64/linux-build/releases/download/0.4.17/xenial-minimal-rock64-0.4.17-85-arm64.img.xz

To reproduce:
1) Open the file above.
2)
```
XZStream image = new XZStream(file);
string uncompressed = sr.ReadToEnd();
```
3) Observe above mentioned exception.

PR #299, opened 2017-09-16
XZ: fix variable-length integers decoding

frabar666:
The `ReadXZInteger` extension method reads VLIs in XZ streams. It reads the right number of bytes, but does not return the correct value for multi-byte VLIs (>= 128) because an increment is missing.

This PR adds the missing increment, with a unit test (based on XZIndex). The fix matches the [reference Java implementation](https://git.tukaani.org/?p=xz-java.git;a=blob;f=src/org/tukaani/xz/common/DecoderUtil.java;h=77ba4413e71ef37f73b29d8dca96231c1c75240a;hb=HEAD#l106).

I see two cases where a valid XZ file could previously not be read:
- XZ file with an index containing >= 128 records (AFAIK indexes aren't that common),
- block filter with >= 128 bytes of properties (seems unlikely).

### DIFF
--- a/src/SharpCompress/Compressors/Xz/MultiByteIntegers.cs
+++ b/src/SharpCompress/Compressors/Xz/MultiByteIntegers.cs
@@ -18,7 +18,7 @@ namespace SharpCompress.Compressors.Xz
             int i = 0;
             while ((LastByte & 0x80) != 0)
             {
-                if (i >= MaxBytes)
+                if (++i >= MaxBytes)
                     throw new InvalidDataException();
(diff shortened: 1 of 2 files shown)


adamhathcock:
I wonder if this fixes https://github.com/adamhathcock/sharpcompress/issues/298
Or
https://github.com/adamhathcock/sharpcompress/issues/296
Or
https://github.com/adamhathcock/sharpcompress/issues/289

I won’t be able to test for a few days.  Thanks for this.

PR #301, opened 2017-09-17
Zlib: fix Adler32 implementation

### LINKED ISSUE #298: ZlibStream fails to check checksum correctly

frabar666:
`ZlibStream`'s `Adler32` routine sometime returns incorrect results, as @ied206 noticed in #298.
This fixes #298 by using `uint` instead of `int`.

There are no unit tests, because the implementation is internal.
[PHP's short test cases](https://github.com/php/php-src/blob/master/ext/hash/tests/adler32.phpt) are OK, but they were already OK before.

My gut says the issue was due to the `%` operator not behaving as expected for negative `int`s, so I tested a 5000-byte 0xFF sequence (to approach the overflow limit at 5552). It exhibits the same pattern as #298, 1st and 5th-8th digits match, but 2nd-4th digits are different: previous implementation returned `0x53e07596`, new one returns `0x54b27596` (same as an online Adler32 calculator).


adamhathcock:
Thanks for this!

PR #313, opened 2017-10-20
bugfix: eliminate spurious rar crc exception

coderb:
eliminate spurious rar crc exception when Read() is called with count = 0


adamhathcock:
Thanks!

Issue #208, opened 2017-02-05
Can't extract encrypted 7z LZMA2 files


CiriousJoker:
I have an encrypted 7zip archive (linked below) encrypted with the password "dontgoout", but whenever I try to extract it, it throws an error at MoveToNextEntry.

Here's a screenshot of the stack trace:
https://gyazo.com/6fb0b0c0731d6e23d04a26e301cc2dbe

I created the archive using the 7zip client for Windows using these settings:
https://gyazo.com/ca4e67e6a259d79cdf0675dc42fd8268

Original name was 0.tmp, but Github only allows zip files.
[0.tmp.encrypted.zip](https://github.com/adamhathcock/sharpcompress/files/752673/0.tmp.encrypted.zip)

Extracting the same archive without any password works perfectly fine and I can extract the encrypted 7z archive using the 7zip desktop client, so this has to be a bug in the library (Or I'm just too dumb for life and I missed something I should have done)

Here's the code I used to extract that archive:

```
SharpCompress.Readers.ReaderOptions test = new SharpCompress.Readers.ReaderOptions();
test.LookForHeader = true;
test.LeaveStreamOpen = false;
test.Password = "dontgoout";

var archive = SevenZipArchive.Open(filename, test);

if (!archive.IsComplete)
{
    MessageBox.Show("One of the downloaded resources is corrupt, please try again.");
    return;
}

long totalbytes = archive.TotalUncompressSize;
long processedbytes = 0;

var reader = archive.ExtractAllEntries();

while (reader.MoveToNextEntry()) // <-- Here it crashes
{
    if (!reader.Entry.IsDirectory)
    {
```

adamhathcock:
I was about to say "sharpcompress doesn't support encrypted 7zip files" but honestly I can't remember.

However, the error you got is fixable because it's using a stream that doesn't support seeking though 7zip requires seeking.

CiriousJoker:
But if the library requires seekable streams (which totally makes sense the way 7z works), why did it work with unencrypted files? I mean they'd have to be seekable too, right?

CiriousJoker:
Woops

adamhathcock:
It must be that the decryption code just needs to know the length for some reason and it's using a different stream class.

The class with the error is one I wrote for the purpose of ensuring no seekable stream was used in other parts of the code.

I haven't done much with the 7zip code personally.  I really hate the 7zip archive format.

CiriousJoker:
[...]

PR #324, opened 2017-12-03
Mk/add support for extracting password protected LZMA(2) 7z archives

CharlieBP:
Added initial support for Password protected 7z (LZMA / LZMA2) archives and added unit tests to confirm the working.


CharlieBP:
Possibly resolves #208

adamhathcock:
Looks great!  Thanks.

michalpenka:
I have a few problems with the archive attached, password "password", created with 7zip 16.04. Had to zip it first to be able to upload it here:
[LZMA.AES.7z.zip](https://github.com/adamhathcock/sharpcompress/files/1855407/LZMA.AES.7z.zip)

Firstly, there is no way to test if archive is encrypted or not. The IsEncrypted property of all the entries is false. To explain a business case - often we get a list of 7z archives and a list of passwords - we need to try the passwords if needed.

Secondly, with password "fdsa" the first two file entries are "successfully" extracted to stream with no exception thrown. The content of these files is, however, still somehow encrypted. The third entry throws an exception.

Thirdly, the invalid password exception would be appreciated. Currently there is no way to catch SharpCompress exceptions as they are protected.

Finally, I have managed to work around the issues mentioned above by trying to extract all the entries into the temp stream first and catching any kind of exception. If no exception is thrown, I know I have a good password (if any). This is of course not ideal as we run through the extraction twice. Also, there is a risk that no exception will be thrown as observed with the first two entries.

Any comments/suggestions/fixes are appreciated.

Thanks for your efforts to support 7z.

adamhathcock:
Sounds like there could be some reworking of encryption handling.  I think creating an issue is more valid here.

I need to check the zip spec but sometimes (maybe just RAR) you only know if the decryption worked if there's an exception.  No other way to know if a password is valid.

me-hungry:
Any updates about if .7z is encrypted or not?

PR #327, opened 2017-12-07
Remove extra padding.  Not sure why I have it but I feel like I did i…

adamhathcock:
…t on purpose


4ybaka:
what's the purpose of writing one empty block? As I understand it should be 2 or none of them.
>> The end of the archive is indicated by two records consisting entirely of zero bytes.

adamhathcock:
I have to review the spec.  I don't really remember.

If you could give me a pull request instead, that would be great.

adamhathcock:
@cyr could you update this as well please?  thanks!

adamhathcock:
Oh wait, this isn't from you @cyr 

I'm way out of date.  I'll update in another PR to remove the extra padding issue

adamhathcock:
Nope...tired again...this isn't needed

PR #331, opened 2017-12-15
Absorb arraypool from CoreFX

adamhathcock:


yinyue200:
Why absorb arraypool

adamhathcock:
Too many different platforms to mistarget.  Easier to do this.

PR #336, opened 2018-01-02
Utility.Skip uses seek

diontools:
Hello.
I have a large zip file.(~100GB 871kEntries)
I tried below code.

```csharp
var sw = Stopwatch.StartNew();
using (var stream = new FileStream(@"large.zip", FileMode.Open, FileAccess.Read, FileShare.Read))
using (var reader = ZipReader.Open(stream, new SharpCompress.Readers.ReaderOptions { LeaveStreamOpen = true }))
{
    {
        //Console.WriteLine(reader.Entry.Key);
    }
}
Console.WriteLine(sw.Elapsed);
```

It took about 13 minutes in my machine.
However, when this PR is applied, it was shortened to 10 seconds.
Does this cause problems?

I'm sorry for my poor English.
Thank you for reading.

### DIFF
--- a/src/SharpCompress/Utility.cs
+++ b/src/SharpCompress/Utility.cs
@@ -141,6 +141,12 @@ namespace SharpCompress
 
         public static void Skip(this Stream source, long advanceAmount)
         {
+            if (source.CanSeek)
+            {
+                source.Position += advanceAmount;
+                return;
+            }
+
             byte[] buffer = GetTransferByteArray();
             try
             {


adamhathcock:
Use the Archive API for random access.  The Reader API intentionally doesn’t use Seek

simmotech:
The change _conditionally_ uses CanSeek to perform a Skip much quicker where possible - why is this considered wrong?

diontools:
Thank you for reply.
I understood Reader spec.

Archive API is very useful, but since Entry is cached, memory space is needed.
In my file it will be about 400 MB.
I'd like to reduce a little more.

I'm trying to extract some files from Zip.
I hope a way to filter Entry.
For example, it is like this.

```csharp
using (var zip = ZipArchive.Open(stream, ...))
{
    foreach (var entry in zip.LoadEntries())
    {
        if (IsTargetFile(entry.Key))
        {
            entry.WriteToDirectory(...);
        }
    }
}
```

Current ZipArchive.LoadEntries is protected, but I think that it would be nice if there was an API similar to this.
Thank you for reading.

adamhathcock:
Fair enough.  I shouldn’t be so rigid.  Hopefully no streams throw when accessing CanSeek

diontools:
Thank you so much. PR closing.

PR #341, opened 2018-01-16
Added ArchiveComment Property to ZipArchive

dos-ise:
Added ArchiveComment Property to ZipArchive

### DIFF
--- a/src/SharpCompress/Common/Zip/SeekableZipHeaderFactory.cs
+++ b/src/SharpCompress/Common/Zip/SeekableZipHeaderFactory.cs
@@ -24,7 +24,7 @@ namespace SharpCompress.Common.Zip
             SeekBackToHeader(stream, reader, DIRECTORY_END_HEADER_BYTES);
             var entry = new DirectoryEndHeader();
             entry.Read(reader);
-
+            var comm = new ArchiveEncoding().Decode(entry.Comment);
             if (entry.IsZip64)
             {
                 _zip64 = true;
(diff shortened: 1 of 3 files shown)


src/SharpCompress/Common/Zip/SeekableZipHeaderFactory.cs
```diff
@@ -24,7 +24,7 @@ internal IEnumerable<DirectoryEntryHeader> ReadSeekableHeader(Stream stream)
             SeekBackToHeader(stream, reader, DIRECTORY_END_HEADER_BYTES);
             var entry = new DirectoryEndHeader();
             entry.Read(reader);
-
+            var comm = new ArchiveEncoding().Decode(entry.Comment);
```
adamhathcock: This isn't going to work properly.  The instance of the encoding object needs to be passed from options.   Maybe don't worry about decoding it yet.

dos-ise: Ok, i moved the decoding to ZipArchive.  

I didn't want to change the type of Comment property on DirectoryEntryHeader. 
So i converted the hex array to string and back.

adamhathcock:
Now that I look at the code here.  Why is this line insufficient? https://github.com/adamhathcock/sharpcompress/blob/f893c1272cbf234036a2c0d693856ed5220cd492/src/SharpCompress/Common/Zip/Headers/DirectoryEntryHeader.cs#L37

I'm already decoding the comment there.  Why are you turning that into a byte array and back?

dos-ise:
Because it does not work. Comment is string.empty after DirectoryEntryHeader.Read() has been executed

The added test proves this. 

The property Comment is string on DirectoryEntryHeader. So the only way to get the decoding from SeekableZipHeaderFactory to ZipArchive is to save it temporarliy in the already existing property. 

Or do you have an better idea?

PR #342, opened 2018-01-18
Add Folders to ZipArchive

dos-ise:
Add Folders to ZipArchive like in dotnetzip. Not yet ready


dos-ise:
You mean the AddAllFromDirectory Extension? 
This should create an empty folder in the zip file instead adding an folder and all of its contetn. 

Its a behavior from dotnetzip we use quit alot.

adamhathcock:
I couldn't quite tell what you were going for.  So you want the ability to add directory entries?

dos-ise:
Correct, I added a test to clarify for what I am trying to achieve. Sorry for the misleading commit.

Issue #86, opened 2015-09-03
ZipReader/StreamingZipReaderFactory fails for archive entries which are uncompressed files in ZIP format


elgonzo:
The issue has already been documented on CodePlex ([see here](https://sharpcompress.codeplex.com/workitem/57)). I added my comments and an example ZIP file exhibiting the problem to this issue in CodePlex, trying to keep the information in place.

A short summary here:
If a ZIP archive contains an uncompressed file entry which happens to be file in the ZIP format (such as another ZIP archive, epub, docx, xlsx document, etc...) then trying to extract the ZIP archive using **ReaderFactory.Open(...)** and **IReader.OpenEntryStream(...)** will fail.

I am using the SharpCompress version 0.11.1 available on nuget.

adamhathcock:
I totally have forgotten about that issue as it wasn't on github.  I'll look into it when I get some free time.

adamhathcock:
I looked at this today and this is basic cause: https://github.com/adamhathcock/sharpcompress/blob/master/SharpCompress/Common/Zip/StreamingZipHeaderFactory.cs#L50

However, this was to fix an issue with zero byte files in a zip file.  I'm not sure I like that fix anyway as it buffers in memory.

Need to think more.

matt-richardson:
Have you had a chance to think about the solution to this issue, @adamhathcock? We've got a customer (of Octopus Deploy) hitting extraction issues with a nested zip file.

The sample file they provided as available at http://help.octopusdeploy.com/discussions/problems/48574, if you want to take a look.

Thanks in advance, and thanks for a great library!

adamhathcock:
The basic issue is that for the Streaming Reader case, it's trying to read all entries.  To do that, it's scanning the file for something that matches a ZIP file header.  

There are possibly two ZIP headers to read: there primary header that's before the data (which may not have a length if it's created using a POST header) and the POST header.  POST headers are created when ZIPs are created with a streaming write to avoid knowing the entire file size.

If the file is being skipped and I don't know the length (because the primary header has no length) then I have to just scan bytes until the next file is found.  Since there's no compression, then I can find a nested ZIP.

I'm not sure this is a solvable problem because in the POST-header no-compression scenario, a ZipReader doesn't know the current file length ever.  So it doesn't know there's a nested ZIP.

Two fixes for this:
- Don't use no-compression ZIPs with a ZipWriter.
- Don't use ZipReader when there's a possibility the above was used.

[...]

PR #344, opened 2018-01-29
Fix Ignoring archive files after processing of internal archive 

### LINKED ISSUE #343: Ignoring archive files after processing of internal archive
Consider archive [ArchiveWithArchive.zip](https://github.com/adamhathcock/sharpcompress/files/1663764/ArchiveWithArchive.zip) with `repositories.xml.zip` volume and `m1.m` inside. `repositories.xml.zip` in its turn contains `repositories.xml` inside.

During processing, SharpCompress extract the internal archive `repositories.xml.zip` with file  `repositories.xml`. But after that library ignores further input archive processing, namely file `m1.m`.

All files should be processed.

codeFather2:
Fix #343


KvanTTT:
@adamhathcock when are you going to merge this PR?
@codeFather2 could you resolve merge conflicts?

adamhathcock:
This is really about non-compressed internal archives which is same as https://github.com/adamhathcock/sharpcompress/issues/86 I believe

codeFather2:
in this case, the problem was that if there is a file in the child archive, the size of which without compression is longer than the remaining length of the parent archive, then after reading the header of this file, the Reader moved to the end of the parent archive, skipping subsequent files. This is due to the fact that when compressing the file header from the child archive, it was placed before the header of the child archive itself, and then it was duplicated inside. My corrections allowed me to bypass this error, without damage to the usual archives

PR #349, opened 2018-02-01
Extension method WriteAll(this IWriter, DirectoryInfo, string, SearchOption, bool)

SimonCahill:
### Extension method WriteAll (see title) with option to prepend directory name.

This extension method allows Sharpcompress to behave similar to the tar executable.

Only available with > net3.5.

### Usage:
To prepend the directory name to the paths within the archive:
```csharp
    tar.WriteAll(archive, option: SearchOption.AllDirectories, prependDirectoryName: true);
```


adamhathcock:
why isn't this just an addition to the previous `WriteAll` ?

Beatsleigher:
I honestly didn't check all the source code. I merely followed the path my code went and looked from there. I didn't want to risk breaking anything. 

But it's a relatively quick and easy change to make.

drewfreyling:
Should this have some tests?

PR #352, opened 2018-02-27
Cake 0.26

adamhathcock:


adamhathcock:
Oh yeah.  I also build and test on .NET Core 1 and 2.  Should I not bother?

PR #354, opened 2018-02-28
Support Deflate64 decompression

frabar666:
This adds support for the Deflate64 decompression algorithm, and its use in ZIP archives. Compression is not implemented.

Deflate64 (method 9 in ZIP archives) is a variation of the standard Deflate algorithm.
In Windows Explorer, compressing >2GB files generates a Deflate64-compressed ZIP file.
7-zip can read and write Deflate64 ZIP files.

Entry point is `SharpCompress.Compressors.Deflate64.Deflate64Stream`, with the same interface as the existing `DeflateStream` (except only decompression is supported).

Implementation is copied from CoreFX, which is MIT-licensed too. Unfortunately it has to be copied, because it is 1. internal, and 2. not supported on all targeted platforms.
Differences between Deflate and Deflate64 are minor, so I first tried to add Deflate64 support in the existing `DeflateStream`... but I failed, hence this distinct implementation.


adamhathcock:
Looks great!

I had thought about using all of the Deflate implementation from .NET Core but yeah it's all internal :(

PR #363, opened 2018-03-27
ZipArchive Reader: Uses IBM PC character encoding to decode filename …

sridhar6668:
According to the zip file format specification, if general purpose bit 11 is unset, the file name and comment should conform to the original ZIP character encoding(IBM PC character encoding set, commonly referred to as IBM Code Page 437.).  If general purpose bit 11 is set, the filename and comment must support UTF8.

https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT . Heading - Language Encoding

Currently, the library always uses UTF8 to decode names. This pull request will use IBM PC character encoding if the UTF8 flag is not set in the header


adamhathcock:
Thanks for this.  I probably just never properly groked the spec.

I think the better change would be to put that if statement and decode the Name directly on the line that does this: `Name = ArchiveEncoding.Decode(name);`

Changing the Decoder on the `ArchiveEncoding` object seems like there would be bigger side-effects.

sridhar6668:
I have updated the code to decode the name directly instead of changing the Decoder on the ArchiveEncoding object

adamhathcock:
The tests fail on this.

Issue #380, opened 2018-05-14
Can't use custom decoder or default encoding for ZIP without UTF8 supporting


saragulov:
After fix https://github.com/adamhathcock/sharpcompress/pull/363 it became impossible to use custom decoder or default encoding for ZIP with null bit 11. However, archives created by Windows 10 tools do not use the IBM 437 encoding. It looks like they use the current user encoding. Previously, you could specify it as the default encoding and the names were decoded successfully. But now it doesn't work.

saragulov:
[Example.zip](https://github.com/adamhathcock/sharpcompress/files/2001325/Example.zip)
See this example zip archive.

adamhathcock:
You're right....I will fix this now.

adamhathcock:
@saragulov look at the PR and see if this works for you...I can merge and do a bugfix release if so.  I believe using Forced to override seems correct.

saragulov:
Thank you for fix again. Can you do a bugfix release?

saragulov:
Thank you very much! Been waiting for the RAR5 support

PR #381, opened 2018-05-14
Allow forced encoding to override default encoding

adamhathcock:
Fixes https://github.com/adamhathcock/sharpcompress/issues/380

### DIFF
--- a/src/SharpCompress/Common/ArchiveEncoding.cs
+++ b/src/SharpCompress/Common/ArchiveEncoding.cs
@@ -49,8 +49,8 @@ namespace SharpCompress.Common
 #if NETSTANDARD1_0
             return Decode(bytes, 0, bytes.Length);
 #else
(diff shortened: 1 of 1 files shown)


saragulov:
Forced encoding does not use in Decode()

adamhathcock:
Oh wait...you're right.  Sorry.  Reworking

adamhathcock:
@saragulov changed again

Trying to rush in between work isn't good

saragulov:
Yes, it works. But i think you should use CustomDecoder check too (along with Forced). They have similar comments ("...for all encoding operations") and CustomDecoder has even higher priority in the GetDecoder method(). Someone can use it with automatic encoding detection for ZIP

adamhathcock:
Yup, you're right.

saragulov:
Thanks!

PR #383, opened 2018-05-23
add example for custom file encoding

itn3000:


adamhathcock:
Thanks!  Samples are very welcome!

PR #384, opened 2018-05-27
Implemented ReadByte/WriteByte on streams to improve performance

MrJul:
`NonDisposingStream` and `ReadOnlySubStream` should implement `ReadByte`: currently they don't, and [the default `Stream` implementation allocates a new one-byte buffer each time](https://source.dot.net/#System.Private.CoreLib/src/System/IO/Stream.cs,757).

`Stream.ReadByte()` is used extensively by the LZMA decoder. With this simple change, I get a ~16-17% speed improvement on my machine while extracting all files in a 43MB LZMA-compressed zip file, and total memory allocations are cut by 1GB!

In the same fashion, I've also overridden `CountingWritableSubStream.WriteByte` to improve encoding. The speed results are marginal (2-3% improvement) since the encoding algorithm takes most of the CPU time, but they exist. Additional, memory traffic is also cut by a fair margin.

I've also implemented `ReadByte`/`WriteByte` on most SharpCompress streams where it was trivial to do so.

### DIFF
--- a/src/SharpCompress/Common/EntryStream.cs
+++ b/src/SharpCompress/Common/EntryStream.cs
@@ -66,6 +66,16 @@ namespace SharpCompress.Common
             return read;
         }
 
+        public override int ReadByte()
+        {
+            int value = _stream.ReadByte();
+            if (value == -1)
+            {
+                _completed = true;
+            }
+            return value;
+        }
+
         public override long Seek(long offset, SeekOrigin origin)
         {
             throw new NotSupportedException();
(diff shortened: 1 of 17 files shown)


adamhathcock:
Thanks for this.  I've barely done any kind of perf or memory usage testing.

If you want to keep doing it and/or implement some kind [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) tests I wouldn't say no :)

SuperJMN:
Looking forward to having it integrated into the NuGet package!

PR #385, opened 2018-05-30
Fixed EFS flag handling

twirpx:
Fixed ArchiveEncoding to use CP437 as default encoding and UTF8 in case when Language encoding flag (EFS) is set.

Fixed DirectoryEntryHeader Name/Comment decoding to use ArchiveEncoding's Forced/Default/CustomEncoder if EFS flag is not set.

Tests fixed to use explicit use of 866 encoding because of file named in Russian usage in several tests.


twirpx:
Tests fails with 'The process cannot access the file 'Tar.tar' because it is being used by another process.'-like error.

Should it be fixed to be friendly to run in parallel?

adamhathcock:
Pondering what the changes to `ArchiveEncoding` do

The tests are unreliable on AppVeyor and CircleCI often.  I believe it's because of virtual file systems.

twirpx:
Please let me know if any changes i have made need explanation.

PR #400, opened 2018-07-10
Avoid throwing NotSupportedException in ReaderFactory hot path

MattKotsenas:
`ReaderFactory.Open()` calls `ZipArchive.IsZipFile()` to determine if the stream is a zip archive, which calls into `ZipHeaderFactory.ReadHeader()`, which throws a `NotSupportedException` when the bytes do not match a valid header.

To be clear, this exception is caught and `IsZipFile()` returns `false`, but when called in a hot-path, these exceptions can become expensive.

To address this issue, `ReadHeader` now returns `null` in the default cause instead of throwing. All callsites were already checking for and handling `null`, so no behavior should change.


adamhathcock:
Thanks, this does make more sense.

PR #403, opened 2018-07-10
Fix and re-enable tests

MattKotsenas:
Fix two problems with the test harness

1. Give each test its own unique scratch space to prevent conflicts between tests

As part of having a unique scratch space, we can remove the `ResetScratch()` method and make the tests simpler and more resilient.

2. Avoid leaking file handles by forcing a garbage collection prior to cleanup for each test

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -77,7 +77,7 @@ Task("Pack")
 Task("Default")
     .IsDependentOn("Restore")
     .IsDependentOn("Build")
-    //.IsDependentOn("Test")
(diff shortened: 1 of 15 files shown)


MattKotsenas:
I have another PR in the works to address the leaking file handles problem, but I wanted to keep these two changes separated, since enabling the tests has value on its own.

adamhathcock:
I appreciate you trying to fix this problem.  I should have thought of doing this which makes me think you came up with a good out of the box solution :)

Issue #237, opened 2017-05-22
Why does AppVeyor hate the tests?


adamhathcock:

adamhathcock:
Should probably just put a build on Travis or Circle that runs the tests on Linux.  They always pass locally on Windows and OS X for me.

adamhathcock:
WTF.  Travis hates the tests too.  Something about the files being stored in Git?

Issue #58, opened 2015-05-21
Concurrency bug in testing


KOLANICH:
When you run tests first time, some tests fail. Then if you rerun unpassed tests, they will pass. Seems like a bug in concurrency

adamhathcock:
I had a static lock in the tests to fix that.  Each TestInitialize would lock and TestCleanup would unlock.

Supposedly, you can get MSTest to not run concurrently as well but I was too lazy to look into it.

PR #404, opened 2018-07-12
Enable parallel test execution

MattKotsenas:
Enable test execution by fixing two classes of file locking problems:

1. Improper use of `IDisposable` in the library
2. Locked file streams by the tests themselves

Solving these problems generally had one of two solutions:

1. Using the `NonDisposingStream` in additional places so types that were "borrowing" a stream could still follow the `Dispose` pattern without upsetting code analysis
2. Changing tests to not take exclusive locks on original files

Lastly, the `IsLocked` check in `TestBase` had to be removed because it isn't parallel-test-execution-aware. Some of that gap is closed by the expanded test coverage in `ReaderTests`, though more is certainly welcome. I chose to stop here to avoid making the PR unmanageable.


MattKotsenas:
Fixes #58

MattKotsenas:
Fixes #237

MattKotsenas:
Thank you for your quick response!

PR #417, opened 2018-10-04
Allow Flush on EntryStream

KyotoFox:
I have a `tar`-file that contains another `tar.gz`-file that I want do extract using SharpCompress.
When I am iterating through the outer `tar`-file, and find the `tar.gz`-file I want to extract, I use a CryptoStream that calculates the SHA256-sum of the `tar.gz`-entry while I extract it (using the IReader extension WriteAllToDirectory)

This seems to work without issues, but somtimes (due to CryptoStreams internal implementation), the `tar.gz` EntryStream will throw a `NotSupportedException` when the CryptoStream is being Disposed, because it will sometimes call `Flush()` on the EntryStream ([see here](https://referencesource.microsoft.com/#mscorlib/system/security/cryptography/cryptostream.cs,141)).

From what I can see there is no reason to not allow flushing the EntryStream, as there is nothing to flush. Microsoft also mentions in the [Stream.Flush-documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.flush?view=netframework-4.7.2#System_IO_Stream_Flush) that read-only streams should implement Flush with an empty method. 
This pull request fixes that, and also adds a test that "provokes" this issue (not sure if that is needed or wanted)


adamhathcock:
You're right.  I think there's no reason to throw an exception on Flush.  Empty implementation is harmless.  This is perfect.  Thank you!

PR #421, opened 2018-10-29
Remove unsafe code

jackfagner:
I made some changes to DataConverter to allow compilation without /unsafe flag.

### DIFF
--- a/src/SharpCompress/Converters/DataConverter.cs
+++ b/src/SharpCompress/Converters/DataConverter.cs
@@ -148,9 +148,7 @@ namespace SharpCompress.Converters
 
         static public DataConverter LittleEndian => BitConverter.IsLittleEndian ? Native : SwapConv;
 
-#if NOTNEEDED
         static public DataConverter BigEndian => BitConverter.IsLittleEndian ? SwapConv : Native;
-#endif
 
         static public DataConverter Native { get; } = new CopyConverter();
 
--- a/src/SharpCompress/Converters/DataConverter.cs
+++ b/src/SharpCompress/Converters/DataConverter.cs
@@ -43,16 +43,12 @@ using System;
 
 namespace SharpCompress.Converters
 {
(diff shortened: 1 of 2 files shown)


jackfagner:
@adamhathcock , sorry. I removed BigEndian by mistake. Now it should compile.

adamhathcock:
Leads into my next question: Why do this? :)

DataConverter came from Mono and fit for older full framework builds.  This could be legacy and removed but I just wanted to understand the motivation.

jackfagner:
> Unsafe code can only be used in a full trust environment. For example, in an ASP.NET server, you usually can't use unsafe code, since it would be pretty easy to introduce a vulnerability by buffer overflow. Another example would be clickonce. Or if your application was accessed from a network share. So if you plan to use your code in a variety of deployment scenario, unsafe code is out of the game.

https://softwareengineering.stackexchange.com/questions/178177/why-not-use-unmanaged-safe-code-in-c/178181

jackfagner:
I looked in to the history of SharpCompress and DataConverter.
Back in 2016, the non-unsafe version of DataConverter was removed:
https://github.com/adamhathcock/sharpcompress/commit/8367bff9fe1e9a89bf33ae1ec5682d2a0353c9c7
https://github.com/adamhathcock/sharpcompress/tree/ff22cd774e2930c1fefd1037a41df756e2f8705a/SharpCompress/Converter
It might be better to re-add DataConverter.Portable.cs and allow non-unsafe compilation with a flag instead of merging this PR?

adamhathcock:
I guess I don't see unsafe as a huge deal as I thought I did it for perf.  Removing it should be fine if that's a concern.  I think most perf gains are to be done elsewhere anyway.

Re-adding the portable version is probably a good idea.

jackfagner:
I'm closing this PR. The code is available if anyone else wants to compile without /unsafe flag.

PR #422, opened 2018-11-01
Tar symlink support

markfinal:
I've been trying to use SharpCompress in my project, and encountered tar archives containing symlinks to be a problem for extraction, so I've tried to implement support for symlinks.

I'm expecting this will bounce back with rejections initially, as there are two holes that I've not filled:

1. currently only supported on .NET standard 2.0 and above, due to the use of Mono.Posix.NETStandard nuget. I'm so far unable to find an equivalent for older versions.
2. What to do on Windows when extracting a symlink.

Suggestions/recommendations for what to do here would be gratefully received.

I've added a test case, which is my initial use case - a source archive of the MoltenVK project on github, which contains symlinks. With local builds, and this PR, I can now use that archive in my project.


adamhathcock:
Thank you for this.

I'm not 100% sure I'm ready to take that dependency.  Primarily because it's `netstandard20` and people are still using this lib on things that don't support that.  I wish they did.

Maybe refactor the actual writing of the symlinks out of the library and just keep the property available.  This puts the symlink writing into your app code which is less than ideal but it's the best compromise I've got for now.

markfinal:
Thanks for the feedback, Adam. I did wonder if the Mono.Posix.NETStandard nuget would be a show stopper.

I was going to suggest another alternative, of instead invoking the `ln` command directly, but I see that's only available from .NET standard 1.3 as a Nuget package, and isn't in 1.1.

Let me play around with the idea of delegating symlink creation (and comparisons, since that's in the test suite), and I will update this PR.

markfinal:
Actually, closing this PR, and I'll open a new one when ready.

PR #423, opened 2018-11-03
Tar symlink support

markfinal:
Reworked the tar symlink support so that the ExtractionOptions now offers a delegate that an application can implement. Exception is thrown if a symlink is encountered but no delegate set.

Mono.Posix.NETStandard removed from the library, but added to the .NET Core 2.1 test app. Implemented the delegate for testing there.


adamhathcock:
Thanks!

markfinal:
Great, thanks!

What's your criteria for making a release? Just wondering what a rough ETA would be to see this in a published Nuget?

MattKotsenas:
Hi @adamhathcock! I also just ran into this problem, and verified that this feature fixes the issue for me. Would it be possible to get a new release?

adamhathcock:
Sorry.  I'll look at getting a new version out soon, I promise!

adamhathcock:
https://www.nuget.org/packages/sharpcompress/0.23.0 done!

Issue #290, opened 2017-08-22
Read ArchiveComment


dos-ise:
Is there a way to read an archivecomment? 
Found a way to write it but not to read.

dos-ise:
I added a pullrequest for this feature.  

https://github.com/adamhathcock/sharpcompress/pull/341

At the moment i am using the follwoing extension.

`using System.IO;
using System.Text;
using System;
using System.Linq;
using System.Reflection;

using SharpCompress.Archives.Zip;

namespace Knx.Ets.Osprey
{
  public static class ZipArchiveExtension
  {
    private const int MAX_ITERATIONS_FOR_DIRECTORY_HEADER = 4096;

    private const uint DIRECTORY_END_HEADER_BYTES = 0x06054b50;

    public static string Comment(this ZipArchive archive)
    {
      var onlyVolume = archive.Volumes.Single();
      var stream = onlyVolume.GetType().GetProperty("Stream", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public).GetValue(onlyVolume) as Stream;
      BinaryReader reader = new BinaryReader(stream);
      SeekBackToHeader(stream, reader, DIRECTORY_END_HEADER_BYTES);

      var VolumeNumber = reader.ReadUInt16();
      var FirstVolumeWithDirectory = reader.ReadUInt16();
      var TotalNumberOfEntriesInDisk = reader.ReadUInt16();
      var TotalNumberOfEntries = reader.ReadUInt16();
      var DirectorySize = reader.ReadUInt32();
      var DirectoryStartOffsetRelativeToDisk = reader.ReadUInt32();
      var CommentLength = reader.ReadUInt16();
      var comment = reader.ReadBytes(CommentLength);

      var comm = Encoding.UTF8.GetString(comment, 0, comment.Length);
      return comm;
    }

    private static void SeekBackToHeader(Stream stream, BinaryReader reader, uint headerSignature)
    {
      long offset = 0;
      uint signature;
      int iterationCount = 0;
      do
      {
        if ((stream.Length + offset) - 4 < 0)
        {
          throw new Exception("Failed to locate the Zip Header");
        }
        stream.Seek(offset - 4, SeekOrigin.End);
        signature = reader.ReadUInt32();
        offset--;
        iterationCount++;
        if (iterationCount > MAX_ITERATIONS_FOR_DIRECTORY_HEADER)
        {
[...]

PR #436, opened 2019-01-13
Changes to populate ZipArchive.ZipVolumne.Comment

Numpsy:
(Related to #290)

Looking at the code, it looks to me like ZipArchive.LoadEntries already contains support for populating the ZipVolume Comment field from the DirectoryEndHeader, but that doesn't get hit because SeekableZipHeaderFactory.ReadSeekableHeader never seems to return the DirectoryEndHeader, it just breaks when there are no more DirectoryEntryHeaders.

This change changes ReadSeekableHeader to return the DirectoryEndHeader, such that the volume comment is populated.

(i'm trying to work out how this stuff works as i go along, and i'm not quite sure how all the cases for returning/breaking in ReadSeekableHeader should ne handled, but the test case suggests that the simple case of getting the archive header works.)

### DIFF
--- a/src/SharpCompress/Common/Zip/SeekableZipHeaderFactory.cs
+++ b/src/SharpCompress/Common/Zip/SeekableZipHeaderFactory.cs
@@ -17,7 +17,7 @@ namespace SharpCompress.Common.Zip
         {
         }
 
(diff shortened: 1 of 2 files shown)


adamhathcock:
Thanks for this.

I fully admit my code paths aren't 100% as I came at this from a "can I make the entire process non-seeking?" vs being feature complete.

PR #442, opened 2019-03-18
Fix: ArchiveEncoding was ignored in TarWriterOptions

turbolocust:
ArchiveEncoding can now be properly used in e.g. TarHeader.


adamhathcock:
Thanks!

PR #444, opened 2019-03-20
Memory and speed optimization

eugeny-trunin:


adamhathcock:
Looks good!

I'll merge this but maybe look at the byte array pooling I'm trying to do too.

Also, I wish I could use the new memory/span classes but not sure they apply.

PR #452, opened 2019-05-24
Various fixes for ZIP64 writer (zero byte entries, 32 bit where supported)

Lssikkes:
To improve compatibility with older ZIP clients I've modified the directory headers to write the corresponding 32 bit values if values still fit in 32 bits (this will allow them to read as if it's a ZIP32, and able to completely ignore the extra ZIP64 fields in ZIP headers if the decompressed size <= 32 bits)

Another problem fixed in order not to trip up 7-zip/WinRAR, zero byte files need to have compression value 0 (STORED). Now with zero byte files in the archive they will validate correctly.

I've tested these changes on WinRAR, 7-zip and windows explorer .zip (Windows 10 RS5), and using the latest changes don't get any complaints anymore extracting ZIP/ZIP64.


adamhathcock:
Great!  Thank you!

PR #453, opened 2019-05-24
Fix for clients failing on just having a 64 bit offset in ZIP64

Lssikkes:
Sorry Adam, while testing some bigger ZIP64 files I found a case where some archive tools don't accept having just 0xFFFFFFFF only for relative offset. This should fix that by also forcing it for compressed/decompressed size to make sure they read everything from the ZIP64 directory record when anything goes over uint32 max for that specific entry.

(according to the spec it's allowed - so technically the commit before was not wrong, but it's probably better to be on the safe side:
https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
   4.4.16 relative offset of local header: (4 bytes)

       This is the offset from the start of the first disk on
       which this file appears, to where the local header SHOULD
       be found.  If an archive is in ZIP64 format and the value
       in this field is 0xFFFFFFFF, the size will be in the 
       corresponding 8 byte zip64 extended information extra field.)

image

image

### DIFF
--- a/src/SharpCompress/Writers/Zip/ZipCentralDirectoryEntry.cs
+++ b/src/SharpCompress/Writers/Zip/ZipCentralDirectoryEntry.cs
@@ -39,9 +39,9 @@ namespace SharpCompress.Writers.Zip
             var zip64 = zip64_stream || HeaderOffset >= uint.MaxValue;
             var usedCompression = compression;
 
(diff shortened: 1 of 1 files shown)


Lssikkes:
Just checked the ZipReader and it seems SharpCompress actually does the right thing, so no need for changes there..

adamhathcock:
Thanks again

PR #457, opened 2019-06-03
Return a bounded substream when data descriptors are used in seekable zips

### LINKED ISSUE #456: Return bounded substreams when data descriptors are used in seekable zips
At my work, we have to process quite a few zip files whose generation we have very little control over. One of the zips we are receiving uses data descriptors (meaning the local file header lengths are zeroed out) and no compression.

SeekableZipFilePart::CreateBaseStream always returns the original BaseStream, which is the entire zip file.  When compression is used, the algorithms are able to detect the end of their compressed parts and stop processing correctly, but, when no compression is used and the local file header lengths are zeroed (as was the case we have to handle), the entire zip file is returned for the ZipFilePart.  So, for ever entry you ask for in the zip file, you essentially get the entire zip file following the entry's local file header.

What I believe SeekableZipFilePart should be doing is returning a bounded substream that utilizes the central directory information when the data descriptors are being used.  I have coded up an implementation that does just this.  It is very narrowly focused on this exact scenario.

PR will be put up momentarily.

DannyBoyk:
Fixes #456


adamhathcock:
I can't seemingly resolve the conflicts in the browser

DannyBoyk:
Since my other PR also added a test, it was just a conflict in ZipArchiveTests.  I updated my fork, rebased this branch on the new master, resolved the conflict, and forced pushed up the changes.  Should be good to go now.  Thanks, Adam.

PR #472, opened 2019-08-21
Fixes #465 EntryExtractionEnd event fired before the Outstream close

### LINKED ISSUE #465: ExtractionEnd event fired before write stream closes
https://github.com/adamhathcock/sharpcompress/blob/a2bd66ded8c9fb8a1b97db34bcc806bb6c0c9030/src/SharpCompress/Archives/IArchiveEntryExtensions.cs#L37

EntryExtractionEnd event fired before the Output stream closes which is causing error while iterating thorough the extracted files. 

I tried to fix it but not sure if that is correct way.
https://github.com/bipinbaglung/sharpcompress/commit/f450817c63946d8dbd40a3d56ded6fc0060bb2da

bipinbaglung:
Fire Entry Extraction End event after write stream closes.

### DIFF
--- a/src/SharpCompress/Archives/IArchiveEntryExtensions.cs
+++ b/src/SharpCompress/Archives/IArchiveEntryExtensions.cs
@@ -34,7 +34,6 @@ namespace SharpCompress.Archives
                     s.TransferTo(streamToWriteTo);
                 }
             }
-            streamListener.FireEntryExtractionEnd(archiveEntry);
         }
 
 #if !NO_FILE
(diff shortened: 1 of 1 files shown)


adamhathcock:
Ohh I see your issue now.  I'm not sure this is the correct fix though because `WriteTo` no longer works with that event.

I think there might need to be a different event listener scenario for the extensions that do file handling.  The events care about the entries (the source stream) while you seem to care about the destination streams.  I didn't make events for this.  

The temporary fix would be using your own extension methods I think.

PR #478, opened 2019-09-11
Use System.Buffers Nuget package

Bond-009:


adamhathcock:
Please describe why this is necessary for you.  I think I did it my way because of platform targeting issues

Bond-009:
@adamhathcock what issues? I did this because it's best practice not to include libs in source control, but via a package manager (i.e Nuget)

adamhathcock:
That’s great and all but the older platforms didn’t like the multi-targeting.  Build isn’t passing anyway.

If I go netstandard only then this will be a thing to do.  I didn’t take it lightly then choosing to absorbing code instead of a reference

Bond-009:
@adamhathcock Fixed the build

Issue #470, opened 2019-08-20
.NET Core (NET Standard) only


adamhathcock:
As full framework is getting left behind (e.g. with the news of .NET 4.8 and .NET 5) I probably only want to support .NET Standard 2.0 and higher.

`netstandard20` should cover the last version of the full framework fully and can add in .NET Core 2.1+ specific things under `netstandard21` or a `netcoreapp` version.

Probably allows for #112 and #233

adamhathcock:
Support for `net46` kept

PR #482, opened 2019-09-17
NET Standard 1.3 and 2.0 only

adamhathcock:
Removes a lot of the build flags for old platforms

#470


adamhathcock:
Adding `net46` to have explicit full framework support for Windows

PR #483, opened 2019-09-17
Use nameof for param names

Bond-009:


adamhathcock:
Thanks.  This is an older code base and likely needs some more love like this.

PR #484, opened 2019-10-03
Pause building of modified collection. Fixes #256

### LINKED ISSUE #256: Slow performance of AddEntry method for big archives (more then 200K)
Is it possible to get rid of [DoesKeyMatchExisting](https://github.com/adamhathcock/sharpcompress/blob/0.17.0/src/SharpCompress/Archives/AbstractWritableArchive.cs#L93) check for `AddAllFromDirectory` method? It leads to non-linear time complexity.

iCodeIT:
Allow for not rebuilding of modified collection while adding all files under a directory.

### DIFF
--- a/src/SharpCompress/Archives/AbstractWritableArchive.cs
+++ b/src/SharpCompress/Archives/AbstractWritableArchive.cs
@@ -17,6 +17,7 @@ namespace SharpCompress.Archives
 
         private readonly List<TEntry> modifiedEntries = new List<TEntry>();
         private bool hasModifications;
+        private bool shouldRebuildModifiedCollection = true;
 
         internal AbstractWritableArchive(ArchiveType type)
             : base(type)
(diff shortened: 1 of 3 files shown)


src/SharpCompress/Archives/AbstractWritableArchive.cs
```diff
@@ -96,10 +97,22 @@ IArchiveEntry IWritableArchive.AddEntry(string key, Stream source, bool closeStr
             }
             var entry = CreateEntry(key, source, size, modified, closeStream);
             newEntries.Add(entry);
-            RebuildModifiedCollection();
+            if(shouldRebuildModifiedCollection)
+                RebuildModifiedCollection();
             return entry;
         }
 
+        public void PauseInternalEntryUpdates()
```
adamhathcock, line 105: I'd make a IDisposable object that would flip the boolean back to true on Dispose so that way a using statement can be used in the implementation

iCodeIT: I thought about that. Such a pause-object would need to know the instance of the AbstractWritableArchive in order to flip the boolean and to call RebuildModifiedCollection on that instance when the pause-object is disposed.
This requires the user to use the using statement - if the user don't and the user doesn't dispose the pause-object himself, then it will be disposed some time later where the AbstractWritableArchive  might have been disposed. It is difficult to get right.
Therefore I opted for the pattern used in WinForms where SuspendLayout/ResumeLayout is commonly used for performance reasons when larger updates of lists or tree structures are updated.

So I don't see how to do this in a nice way (but I am happy to be educated on the issue ;-) if you can show me how)

adamhathcock:
Check out https://github.com/adamhathcock/sharpcompress/pull/485

PR #485, opened 2019-10-09
Create and using PauseEntryRebuilding for adding large numbers of ent…

adamhathcock:
…ries


adamhathcock:
Fixes 256

Superceeds - https://github.com/adamhathcock/sharpcompress/pull/484

src/SharpCompress/Archives/AbstractWritableArchive.cs
```diff
@@ -12,11 +12,28 @@ public abstract class AbstractWritableArchive<TEntry, TVolume> : AbstractArchive
...
+
+            public RebuildPauseDisposable(AbstractWritableArchive<TEntry, TVolume> archive)
+            {
+                this.archive = archive;
+                archive.pauseRebuilding = true;
+            }
+
+            public void Dispose()
+            {
+                archive.pauseRebuilding = false;
```
iCodeIT, line 27: What happens here if dispose was already called on archive? 
Just because you can use using do this nicely, there is no guarantee people do it. So this is my main concern for using this pattern.

src/SharpCompress/Archives/IWritableArchiveExtensions.cs
```diff
@@ -39,15 +39,21 @@ public static void SaveTo(this IWritableArchive writableArchive, FileInfo fileIn
...
-                writableArchive.AddEntry(path.Substring(filePath.Length), fileInfo.OpenRead(), true, fileInfo.Length,
-                                         fileInfo.LastWriteTime);
+                {
+                    var fileInfo = new FileInfo(path);
+                    writableArchive.AddEntry(path.Substring(filePath.Length),
+                                             fileInfo.OpenRead(),
+                                             true,
+                                             fileInfo.Length,
+                                             fileInfo.LastWriteTime);
+                }
```
iCodeIT, line 45: How about instead of using IDispose we have:
void PauseEntryBuilding(Action<.....> action)
{
 Pause();
 try
 {
   action(...);
 }
 finally
 {
   Resume();
 }
}

called like 

PauseEntryBuilding((....) => 
{
  foreach .... addEntry....
});

adamhathcock: I see what you mean but I don't see that as better.   Using `using` feels more C# like

src/SharpCompress/Archives/AbstractWritableArchive.cs
adamhathcock: Calling dispose multiple times doesn't matter.  If people use `PauseEntryRebuilding` multiple times incorrectly it's not any worse than them using the previous suggestion or the suggestion below.  It's an education issue.

iCodeIT: It is not about multiple calls to Dispose, but that you might reference a disposed object "archive" in this dispose method. That is bad design.

src/SharpCompress/Archives/IWritableArchiveExtensions.cs
iCodeIT: That depends on which version of C# you use - this is a bit functional programming like, but completely normal.

src/SharpCompress/Archives/AbstractWritableArchive.cs
iCodeIT: Also calling "Rebuild-whatever" in Dispose is a misuse of dispose. Dispose is for cleaning up after yourself, not running arbitrary code in another class.

adamhathcock: Archive will never be null.  It doesn’t matter if archive is disposed because this object only ever flips a Boolean on archive.  The design is exploiting the using/Dispose pattern.  It’s a single purpose not meant for anything else.

iCodeIT: The word "exploit" should tell you this is wrong. It goes against recommended and established practices. (If you have a hammer it doesn't mean you should treat everything as a nail). Simply put, this is how you build technical debt.

It is your repository and you can of course do whatever you want - but I would recommend to go with the Action() approach. It is a bit more like functional programming but has been part of C# for many years - and it cleanly does what you want: It ensures that after pause comes a resume with rebuilding of the collection.

PR #495, opened 2019-12-29
Use functions from System.Memory instead of selfmade ones

Bond-009:
Replaces multiple helper functions to fill arrays with `Span.Fill` calls.

Also improved the performance of the `Utility.SetSize` function (and replaced 1 instance)


tests/SharpCompress.Test/SharpCompress.Test.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.2</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
adamhathcock: Please change this back for now.  I haven't updated things across the board to 3.1 yet.

Bond-009: Sorry, had to do this to get it running on my machine.

PR #496, opened 2019-12-30
Reduce the amount of allocations

Bond-009:
* Directly fill an array instead of filling a List and copying that to
an array
* Use own buffer when writing bytes to a stream
* Remove DataConverter class, replaced by BinaryPrimitives


adamhathcock:
Need to bump the lang version on the csproj.  
```
ompressors\ADC\ADCBase.cs(119,31): error CS8107: Feature 'ref structs' is not available in C# 7.0. Please use language version 7.2 or greater. [C:\projects\sharpcompress\src\SharpCompress\SharpCompress.csproj]
```
I should just do a blanket upgrade for building and also maybe a target for `netstandard2.1` as well

Issue #464, opened 2019-08-08
SharpCompress hangs when tar contains symbolic link and is not possible to create it


justalemon:
I was currently working on one of my programs when I tried to extract a `.tar.xz` (you can get it from [here](https://runtime.fivem.net/artifacts/fivem/build_proot_linux/master/1462-ba3c7493bf62c6fab054987761c41d3b6c696889/)) and the WriteEntryToDirectory call ended up hanging on a specific file:

![image](https://user-images.githubusercontent.com/11861253/62694967-5d70eb80-b9a3-11e9-8cf8-000f2681514e.png)

Just to make sure that I didn't made a mistake on my code, I tried to extract the file with WinRAR and it warned me that it was not possible to create a symbolic link:

![image](https://user-images.githubusercontent.com/11861253/62695126-b0e33980-b9a3-11e9-8ec8-dfbc5fb468f0.png)

> Unable to create the symbolic link C:\Users\lemon\AppData\Local\Temp\Rar$DRa13524.1044\alpine\usr\share\terminfo\l\linux
> Maybe you need to execute WinRAR with administrator permissions.

Maybe SharpCompress is not handling symbolic links errors correctly?

justalemon:
Looks like this is related to either Mono or WSL/Ubuntu.

adamhathcock:
Anymore detail?  Maybe this is a simple bug?  I use a MacOS so I haven't tried to reproduce this.

RealOrko:
Just found the same thing using v0.25.0 on Fedora 31.

```
SharpCompress.Common.ExtractionException: Entry is a symbolic link but ExtractionOptions.WriteSymbolicLink delegate is null
   at SharpCompress.Common.ExtractionMethods.WriteEntryToFile(IEntry entry, String destinationFileName, ExtractionOptions options, Action`2 openAndWrite)
```

By settings up the options as follows it stopped throwing an exception: 

```csharp
var opt = new ExtractionOptions {
   ExtractFullPath = true,
   Overwrite = true,
   WriteSymbolicLink = (source, target) => {} 
};
```

After doing a little investigation it would seem there is no clear way of dealing with this in the DOTNET Runtime until [this issue](https://github.com/dotnet/runtime/issues/24271) is solved.

adamhathcock:
Added default delegate in https://www.nuget.org/packages/sharpcompress/0.25.1

RealOrko:
[...]

PR #513, opened 2020-05-21
Added default implementation with warning for symbolic links

RealOrko:
Hi 

I changed the default implementation for writing symbolic links to emit a warning instead of throwing an exception for this [issue](https://github.com/adamhathcock/sharpcompress/issues/464). 

Not sure how this runs for performance. 

Another thing we could try is to create platform specific commands that call out to a shell command for the various platforms. 

Something like:

```csharp
if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) {
   ShellCommands.Exec($"mklink /H {source} {target}")
} else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) {
   ShellCommands.Exec($"ln -s {target} {source}")
} else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
   ShellCommands.Exec($"ln -s {target} {source}")
}
```

### DIFF
--- a/src/SharpCompress/Common/ExtractionOptions.cs
+++ b/src/SharpCompress/Common/ExtractionOptions.cs
@@ -1,4 +1,6 @@
-﻿namespace SharpCompress.Common
+﻿using System;
+
+namespace SharpCompress.Common
 {
     public class ExtractionOptions
     {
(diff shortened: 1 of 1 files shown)


RealOrko:
After using [this source code](https://github.com/RealOrko/dotnet-cli-ghinstaller/tree/master/src/ghinstaller/Modules/Process) for launching OS specific shell commands, I did manage to unpack a tar with symbolic links using this code: 

https://github.com/RealOrko/dotnet-cli-ghinstaller/blob/232d13a57121cb421b6b10d0f3695dddfa840c02/src/ghinstaller/Modules/Archiving/Archive.cs#L29-L54

adamhathcock:
Thanks for this.  Should probably put some kind of warning in.

PR #515, opened 2020-05-23
Enable nullable

iamcarbon:
This PR

- Enables nullable
- Replaces == null with is null
- Seals a few internal classes
- Remove a single allocation by creating a static byte array

Nullable is disabled in various classes with initialization patterns that don't work well with nullable until C# v9.


iamcarbon:
@adamhathcock Ready for review. 

If things look good, I'll do another PR that focused on eliminating allocations. I spotted at least 20 spots that could be optimized but didn't want this PR to get to big.

src/SharpCompress/Archives/AbstractArchive.cs
```diff
@@ -48,12 +48,14 @@ internal AbstractArchive(ArchiveType type, IEnumerable<Stream> streams, ReaderOp
             lazyEntries = new LazyReadOnlyCollection<TEntry>(LoadEntries(Volumes));
         }
 
+#nullable disable
```
adamhathcock, line 51: Not too familiar but why is this necessary?

src/SharpCompress/Archives/GZip/GZipWritableArchiveEntry.cs
```diff
@@ -1,12 +1,14 @@
-﻿using System;
+﻿#nullable disable
```
adamhathcock, line 1: Disabled for file?

src/SharpCompress/Archives/Rar/RarArchiveVolumeFactory.cs
```diff
@@ -34,20 +34,20 @@ internal static IEnumerable<RarVolume> GetParts(FileInfo fileInfo, ReaderOptions
             {
                 yield break; //if file isn't volume then there is no reason to look
             }
-            fileInfo = GetNextFileInfo(ah, part.FileParts.FirstOrDefault() as FileInfoRarFilePart);
+            fileInfo = GetNextFileInfo(ah, part.FileParts.FirstOrDefault() as FileInfoRarFilePart)!;
```
adamhathcock, line 37: Should fileInfo just be nullable?

src/SharpCompress/Archives/Zip/ZipArchive.cs
```diff
@@ -15,7 +15,9 @@ namespace SharpCompress.Archives.Zip
 {
     public class ZipArchive : AbstractWritableArchive<ZipArchiveEntry, ZipVolume>
     {
+#nullable disable
```
adamhathcock, line 18: Why disable for this?  It should never be null

adamhathcock:
I should have read your general disabling comment.  I can't wait for C#9 improving general usage.  I didn't bother turning on nullables in this because it was a lot of work.

PR #530, opened 2020-07-31
Enable test coverage for net461 and fix regression

iamcarbon:
This PR enables test coverage for .NET 4.6.1 and fixes an error introduced in my last PR.

There are 5 failing ZipWriter tests that are currently failing that are conditionally excluded. These have been failing for a while (before any recent PRs) and should be addressed in a separate PR.

### DIFF
--- a/src/SharpCompress/IO/NonDisposingStream.cs
+++ b/src/SharpCompress/IO/NonDisposingStream.cs
@@ -62,7 +62,7 @@ namespace SharpCompress.IO
 
         public override int Read(Span<byte> buffer)
         {
-            Stream.Read(buffer);
+            return Stream.Read(buffer);
         }
 
         public override void Write(ReadOnlySpan<byte> buffer)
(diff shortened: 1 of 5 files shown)


iamcarbon:
@adamhathcock Ready for review. This fixes a bug I introduced in my last PR (ugh).

It looks like the net461 tests are failing due to an encoding issue.

тест.txt is getting replaced with ????.txt

We'll want to address this and re-enable these tests to ensure we have good test coverage over the encoders.

adamhathcock:
I'm torn because I don't want to bother supporting legacy framework once .NET 5 is out :)

I'll probably need to enable the CI to use windows to see the exact failure since I don't use windows.  

I'm guessing the fix is similar to the recent test changes I did to fix the mobile/Blazor encoding issue.

iamcarbon:
Just confirmed that updating the Default ArchiveEncoding to UTF8 (from ASCII) also fixes this.

iamcarbon:
Overwriting the default encoding selectively in the tests worked. All the tests pass on the legacy and modern frameworks.

+ 1 to dropping legacy frameworks after the .NET 5 release.

PR #531, opened 2020-08-01
Improve CQ3

iamcarbon:
This PR eliminates various allocations by utilizing the stack to allocate short lived arrays.


adamhathcock:
Thanks again.  I like reading these as I want to think in terms of Spans instead of byte arrays.

Issue #532, opened 2020-08-05
Index out of range exception from gzip


xy1473210653:
my code:
using(var gzip =File.OpenWrite(dirPath))
using(var write = WriteFactory.Open(gzip,ArchiveType.GZip,new WriterOptions(CompressionType.GZip){LeaveStreamOpen = true}))
{
  write.WriteAll(filePath,"*",SearchOption.AllDirectories);
}

xy1473210653:
Why should compressed packages in. Tar format adopt the compression type of. Gzip

avao:
It looks like there is an issue with extracting file name in GZipStream.cs.
Try extracting file name in your code and supply it to Write (instead of sending full path).

PR #541, opened 2020-10-13
UT and Fix for: Index out of range exception from gzip #532

avao:


src/SharpCompress/Compressors/Deflate/GZipStream.cs
```diff
@@ -384,16 +384,7 @@ public override void Write(byte[] buffer, int offset, int count)
...
                 {
                     // trim any leading path
-                    int length = _fileName.Length;
-                    int num = length;
-                    while (--num >= 0)
-                    {
-                        char c = _fileName[num];
-                        if (c == '\\')
-                        {
-                            _fileName = _fileName.Substring(num + 1, length - num - 1);
```
avao, line 394: the main issue was here - missing break after this line. I decided to simplify code though.

adamhathcock:
Thanks, this is way simplier!

PR #546, opened 2020-11-02
Update .NET SDK

iamcarbon:


iamcarbon:
Need to remove global.json

adamhathcock:
Why remove global.json?  Just update both.  Global.json helps enforce SDK version

PR #565, opened 2021-02-01
On ongoing branch of making everything async

adamhathcock:
- Use ReadAsync/WriteAsync on every stream
- Use CancelationToken (drop Listener interface?)
- Use ValueToken (need to find better Task.Completed version)
- Use IAsyncDisposable (drop IDispose as it mostly doesn't apply)
- Use IAsyncEnumerable
- More?

### DIFF
--- a/src/SharpCompress/Archives/AbstractWritableArchive.cs
+++ b/src/SharpCompress/Archives/AbstractWritableArchive.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using SharpCompress.Common;
 using SharpCompress.Readers;
 using SharpCompress.Writers;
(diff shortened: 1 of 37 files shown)


src/SharpCompress/Archives/AbstractWritableArchive.cs
```diff
@@ -137,11 +139,11 @@ private bool DoesKeyMatchExisting(string key)
             return false;
         }
 
-        public void SaveTo(Stream stream, WriterOptions options)
+        public async Task SaveToAsync(Stream stream, WriterOptions options)
```
JanHyka: CancellationToken parameter should be passed through whole call chain and exposed to library consumers (as non-mandatory parameter on top level). Otherwise caller won't be able to cancel running operation (timeouts).

adamhathcock: It's definitely something I want to do.  I might miss more.  Need to find a Rider plugin to help me catch it.

Ilanlido:
is there any plan on continuing this work?

adamhathcock:
Not from me anytime soon as my personal and professional life is too busy.  Sorry.

BloodShadow174:
Hi. Can we expect asynchrony support in the near future?

PR #568, opened 2021-02-12
Use stackallocs where possible/sensible

Bond-009:


adamhathcock (approved):
Thanks for this.

I'm getting used to Span/Memory along aside making literally everything async here: https://github.com/adamhathcock/sharpcompress/pull/565

One thing I learned: span/stackalloc can't be used in an async method but it's not a big deal.

PR #569, opened 2021-02-13
Improve how missing parts are handled

BrendanGrant:
I have been doing some work involving processing incomplete RAR files and file sets, and I had to make a few tweaks to make SharpCompress ok with trying to read the partial data and figured they useful enough to share.

### DIFF
--- a/src/SharpCompress/Archives/Rar/RarArchive.cs
+++ b/src/SharpCompress/Archives/Rar/RarArchive.cs
@@ -10,7 +10,8 @@ using SharpCompress.Readers.Rar;
 
 namespace SharpCompress.Archives.Rar
 {
-    public class RarArchive : AbstractArchive<RarArchiveEntry, RarVolume>
+    public class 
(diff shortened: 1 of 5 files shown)


adamhathcock (approved):
Always useful to have more options like this.  Thanks!

BrendanGrant:
I'm pleased you like the option, however I didn't expect you to merge it so quick as the tests are failing (I'd never actually run them myself as the changes were so small). One is an easy fix, just changing the expected exception, the others require a good bit more consideration as I think they are making some assumptions which aren't always accurate.

adamhathcock:
Actually the fix was better split detection by default: https://github.com/adamhathcock/sharpcompress/commit/a51b56339a9c221a2bab2e21877c1295a5551c27

I knew things were failing but I like merging in good ideas.  If you're not done, consider using the Draft PR feature.  Can discuss without worrying about a quick merge if you're not ready.

Issue #555, opened 2020-12-13
ZIP files with a comment larger than 4096 throws exception


elephantflea:
If a ZIP file has a comment attached to the archive (not an entry) and it is over 4096 bytes it cannot be extracted as it will throw the exception:

`SharpCompress.Common.ArchiveException: Could not find Zip file Directory at the end of the file.  File may be corrupted.`

 When a call to list all entries is made. eg using [LongComment.zip](https://github.com/adamhathcock/sharpcompress/files/5684438/LongComment.zip)

```csharp
using (var stream = File.OpenRead("LongComment.zip"))
using (var archive = ZipArchive.Open(stream))
{
    foreach(var entry in archive.Entries)
    {
        Console.WriteLine(entry.Key);
    }
}
```
I assume this is because of the constant `MAX_ITERATIONS_FOR_DIRECTORY_HEADER` in `SeekableZipHeaderFactory.cs`, but I would think there is a way to factor in the archive comment length into this logic.

adamhathcock:
Released https://github.com/adamhathcock/sharpcompress/releases/tag/0.28

PR #570, opened 2021-02-13
Propsal for handling Zip with long comment

Erior:
Proposal for fixing #555


adamhathcock (approved):
Looks sound.   Thanks!

Issue #540, opened 2020-10-13
Extra bytes written when set to zip64


brecollog:
I have seen various flavors of this issue, some event claim this issue is fixed but I have downloaded the most recent version of sharpcompress and the issue is still there.  When writing out a large file, i.e. >2GB, and using the zip64 flag set to true, extra bytes get written in the header that cause issues when decompression.  Very repeatable.  Create a single file entry > 2GB and add it to the archive and save. If you have 7zip you can right click on the archive and select 'test archive' and you will get the error.

Does anyone have a fix as to how to combat this issue?  My apologies if I have missed how this has been addressed.

adamhathcock:
As far as I can remember, I don't know if there's been anything specific done to fix this.

The error is extra bytes in the header?  Can you give text or screen shots for details?

brecollog:
Using this code after adding SharpCompress:

using System;
using SharpCompress.Common;
using SharpCompress.Writers;
using  System.IO;
namespace SharpCompressConsoleTest
{
    class Program
    {
        static void Main(string[] args)
        {
            string filename = Path.Combine(@"c:\temp", Guid.NewGuid().ToString() + ".zip");
            using (FileStream fs = new FileStream(filename, FileMode.Create, FileAccess.Write))
            {
                var options = new SharpCompress.Writers.Zip.ZipWriterOptions(CompressionType.Deflate)
                {
                    UseZip64 = true
                };

                using (var writer = WriterFactory.Open(fs, ArchiveType.Zip, options))
                {
                    //dummy.txt is 4GB text file.  it doesn't have to be > 4GB.
                    //this can be a 1kb file.  Simply selecting UseZip64 creates a header error
                    writer.Write(@"c:\temp\abc.txt", new FileInfo(@"c:\temp\abc.txt"));
                }
            }
        }
    }
}
This is simply creating a zip file with the UseZip64 flag set, which you must do when creating files >4GB or sharpcompress will generate an error "System.NotSupportedException: 'Attempted to write a stream that is larger than 4GiB without setting the zip64 option'

This code above will create a zip file but will generate a header error. If you have 7z installed you can right click the file and select 7 Zip/Test Archive and see that it has the header error.

adamhathcock:
Released https://github.com/adamhathcock/sharpcompress/releases/tag/0.28

PR #571, opened 2021-02-14
Proposal fixing Extra bytes written when setting zip64

Erior:
Zip64 requires version 4.5
Number of disks is 4 bytes and not 8

#540


adamhathcock (approved):
Thanks!

Issue #539, opened 2020-10-12
entry.IsEncrypted is false when a 7z file has an AesDecoderStream


Sicos1977:
Is it somehow possible to set the IsEncrypted to true when the 7z file has an AesDecoderStream?

I use the IsEncrypted flag to check is a file is password protected so in this case I would expect it to be true

        internal static Stream CreateDecoderStream(CMethodId id, Stream[] inStreams, byte[] info, IPasswordProvider pass,
                                                   long limit)
        {
            switch (id._id)
            {
                case K_COPY:
                    if (info != null)
                    {
                        throw new NotSupportedException();
                    }
                    return inStreams.Single();
                case K_LZMA:
                case K_LZMA2:
                    return new LzmaStream(info, inStreams.Single(), -1, limit);
                case CMethodId.K_AES_ID:
                    return new AesDecoderStream(inStreams.Single(), info, pass, limit);
                case K_BCJ:

Sicos1977:
Any updates on this?

Sicos1977:
Any updates?

adamhathcock:
Happy to accept pull requests

Sicos1977:
If you can give me some ideas how to do this I would be happy to do it.

Issue #521, opened 2020-06-23
SevenZipEntries.IsEncrypted is always set to False


evpxregu:
When opening a sevenziparchive and checking the entries in the archive for all the entries IsEncrypted will be set to false.

You can see in the sourcode that the value is hardcoded.

https://github.com/adamhathcock/sharpcompress/blob/00ff119ec4f13020fb40bba3654d53e9836ebd6d/src/SharpCompress/Common/SevenZip/SevenZipEntry.cs

Sicos1977:
I have the same problem and made issue #539 for it

PR #572, opened 2021-02-14
Not so elegant perhaps for checking 7z encryption

Erior:
#539
#521 

Could this be a solution?


adamhathcock:
Looks elegant enough to me.  I don't know enough about the 7Zip spec off hand to know if that's correct though.

Erior:
Digging some more I don't think it is too far off

[p7zip discussion on sourceforge](https://sourceforge.net/p/p7zip/discussion/383044/thread/082d0085/)

More complete might be to just detect that it starts with 06 xx xx xx

From 7zip DOC/Methods.txt
06.. - Crypto 

   F0 - Ciphers without hashing algo

      01 - [AES]
         0x - AES-128
         4x - AES-192
         8x - AES-256
         Cx - AES

         x0 - ECB
         x1 - CBC
         x2 - CFB
         x3 - OFB
         x4 - CTR

   F1 - Combine Ciphers

      01 - [Zip]
         01 - ZipCrypto (Main Zip crypto algo)

      03 - [RAR]
         02 - 
         03 - Rar29AES (AES-128 + modified SHA-1)

      07 - [7z]
         01 - 7zAES (AES-256 + SHA-256)

adamhathcock (approved):

PR #573, opened 2021-02-17
Add read-only support for Dmg archives

Artentus:
As the title suggests, this adds support for reading and extracting Dmg archives.  
Only Dmg archives that use the GPT partitioning scheme with a HFS+ or HFSX data partition are supported, but that should cover the large majority of Dmgs out there (any DMGs created on MacOS 8.1 and newer will have this format).

I didn't see any way of implementing this as read-through so only the archive API is available.  
Unfortunately I'm not able to provide test data as I don't have access to a Mac to create Dmgs. I have tested extensively using my own data but I cannot share that data publicly. Hopefully someone else will be kind enough to provide those.


adamhathcock:
This is great!  Thanks!

I have a Mac.  I need to figure out how to make a dmg with test data

adamhathcock:
Maybe I'm not making a valid DMG but AFPS isn't supported and HFS reports no entries:
https://github.com/adamhathcock/sharpcompress/pull/602

PR #579, opened 2021-03-18
Do not place extension classes in common namespace

Looooong:
Do not place extension classes in common namespace, as the extension methods might conflict with the same counterpart (same name, same signature) from another separated namespace.


aisk (approved):

adamhathcock:
Thanks!

PR #583, opened 2021-03-30
Added Seeking feature for BufferedSubStream and ReadOnlySubStream.

vovgou:
Added the seeking feature for BufferedSubStream and ReadOnlySubStream. When the compression type of an entry is Store (uncompressed) in the Zip or APK file, a seekable stream will be obtained by entry.OpenEntryStream(). In game development on the Android platform, it avoids having to decompress files to disk due to a seekable stream being required.

### DIFF
--- a/src/SharpCompress/IO/BufferedSubStream.cs
+++ b/src/SharpCompress/IO/BufferedSubStream.cs
@@ -5,79 +5,247 @@ namespace SharpCompress.IO
 {
     internal class BufferedSubStream : NonDisposingStream
     {
-        private long position;
-        private int cacheOffset;
-        private int cacheLength;
-        private readonly byte[] cache;
+        private const int DEFAULT_BUFFER_SIZE = 32768;
+        private readonly long startIndexInBaseStream;
+        private readonly long endIndexInBaseStream;
+        private long positionInBaseStream;
+        private readonly long length;
+        private readonly Stream baseStream;
+        private bool canRead;
+        private bool disposedValue;
 
-        public BufferedSubStream(Stream stream, long origin, long bytesToRead) : base(stream, throwOnDispose: false)
+        private readonly int bufferSize;
+        private int bufferOffset;
+        private int bufferLength;
+        private byte[] buffer;
+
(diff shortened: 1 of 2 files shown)


vovgou:
Hi，@adamhathcock
My submission failed to pass the test case, but this does not seem to be my problem. I tested some cases in which errors were reported on my computer, and it worked.

adamhathcock:
Not sure how these pass locally for you as some of the code looks at the `CanSeek` flags to do different things.  That's why the test is failing because `ReadOnlySubStream` was never seekable and now `RewindableStream` blows up because it's trying to seek.

I believe `entry.OpenEntryStream()` should always return a non-seekable stream as it's directly linked to decompression and inheritly shouldn't be seekable.  If you want to buffer it for your app, do so, but I don't think this is something that should be part of SharpCompress.

vovgou:
> Not sure how these pass locally for you as some of the code looks at the `CanSeek` flags to do different things. That's why the test is failing because `ReadOnlySubStream` was never seekable and now `RewindableStream` blows up because it's trying to seek.
> 
> I believe `entry.OpenEntryStream()` should always return a non-seekable stream as it's directly linked to decompression and inheritly shouldn't be seekable. If you want to buffer it for your app, do so, but I don't think this is something that should be part of SharpCompress.

Usually, entry.OpenEntryStream() gets a compressed stream that cannot be seeked, but in the zip file, if the Entry is a "Store" type, it is not compressed. In the game developed with Unity3d (Android APK), The files in the StreamingAssets directory are game resource files. They are all "Store" types and are not compressed. It is necessary to support a seekable stream. And when using SharpZipLib, it works.So could you support this feature? Thanks you.

adamhathcock:
I guess if there's no decompression stream, then it could be a seekable stream when using OpenEntryStream.  Just need to complete the thread and check for uncompressed streams in zips

vovgou:
> I guess if there's no decompression stream, then it could be a seekable stream when using OpenEntryStream. Just need to complete the thread and check for uncompressed streams in zips

Yes, it is seekable, but it is wrapped by ReadOnlySubStream in sharpcompress, so i added the seeking feature to ReadOnlySubStream

Issue #595, opened 2021-04-26
Zip64 Extra Header is not parsed conditionally


DannyBoyk:
The current logic for the Zip64 extra header assumes that certain fields are present by the number of bytes in the extra header, but the specification indicates that the fields that are being parsed are based on the values in the local entry header, namely that the fields in the extra header correspond to those fields in the local header that are set to uint.MaxValue (or ushort.MaxValue for the disk number).

We have encountered Zip64 files in our production systems that only specify a compressed size in their Zip64 extra fields and SharpCompress assumes these represent the uncompressed size since only one value is specified.

adamhathcock:
Released https://www.nuget.org/packages/SharpCompress/0.28.3

PR #596, opened 2021-04-26
Conditionally parse Zip64 extra field based on specification

DannyBoyk:
The Zip64 extra field should look for values based on the corresponding
values in the local entry header.

Fixes adamhathcock/sharpcompress#595


runemoennike:
@adamhathcock any plans for merging this into an upcoming release? :) We would love to have this fix from the nuget package instead of changing our dependency.

adamhathcock:
Thanks for this!

PR #601, opened 2021-05-23
Write ZIP64 End of Central Directory only if needed.

salvois:
Both Excel and LibreOffice fail to open an XLSX file compressed with SharpCompress when using ZIP64 (Excel is able to repair it, LibreOffice refuses it altogether).

Avoiding the ZIP64 End of Central Directory record when values are representable by the non-ZIP64 End of Central Directory record solves the issue.

This seems to be supported by the ZIP file format specification: "If one of the fields in the end of central directory record is too small to hold required data, the field SHOULD be set to -1 (0xFFFF or 0xFFFFFFFF) and the ZIP64 format record SHOULD be created." (https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT version 6.3.9, paragraph 4.4.1.4).


adamhathcock:
Seems correct.  Thanks.

adamhathcock:
Released https://www.nuget.org/packages/SharpCompress/0.28.3

PR #610, opened 2021-08-27
Bugfix for TarWriter - too much padding in large files

cyr:
For large files, due to how the padding was calculated an additional 512 bytes was added. I noticed this while archiving a file (size 11721453867 bytes, i.e. 11gb, roughly) and 725 null bytes were added as padding, instead of 213. This new method calculates large sizes correctly (and is faster!).


adamhathcock:
Could you update the PR with the latest master to make this mergable?

cyr:
Done!

adamhathcock:
Thanks!

adamhathcock:
Released in https://github.com/adamhathcock/sharpcompress/releases/tag/0.29.0

PR #611, opened 2021-09-02
Allowing to seek empty zip files

Thunderstr1k3:
In our workflow it may happen that we attempt to access zip files that don't actually have any content. Until earlier versions that worked, but with the last implementation of SeekableZipHeaderFactory it fails. I suggest parsing the entire stream instead of skipping the oecd region. Respective unit test is added, too.

If you accept the changes, please publish a new nuget package.

Best regards

### DIFF
--- a/src/SharpCompress/Common/Zip/SeekableZipHeaderFactory.cs
+++ b/src/SharpCompress/Common/Zip/SeekableZipHeaderFactory.cs
@@ -117,7 +117,8 @@ namespace SharpCompress.Common.Zip
             // Search in reverse
             Array.Reverse(seek);
 
-            var max_search_area = len - MINIMUM_EOCD_LENGTH;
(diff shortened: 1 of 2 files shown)


bormm:
Hi @adamhathcock, I am the team lead of the Foxit Europe Enterprise Solutions cloud team. We are hanging because of this problem. OpenSource should be not be only on consuming but also on giving back, so my team decided to not just "report the issue" but provide a PR to fix it here. Is there anything development related that can be done by my team to support you merging this request and providing a new nupkg? Like extending the automatic checks here or something?

adamhathcock:
I only just noticed this.  It smells okay and thanks for providing a test!

adamhathcock:
Released in https://github.com/adamhathcock/sharpcompress/releases/tag/0.29.0

Issue #591, opened 2021-04-13
SharpCompress 0.28.1 depends on System.Buffers 4.4.0, but NuGet transitively depends on 4.5.1


schlotter:
# Bug description
Using SharpCompress 0.28.1 in a .NET 4.6.2 unit test project I get the following exception:
```
System.IO.FileLoadException: Could not load file or assembly 'System.Buffers, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
    at SharpCompress.Utility.Skip(Stream source)
   at SharpCompress.Common.EntryStream.Dispose(Boolean disposing)
   at System.IO.Stream.Close()
   at System.IO.Stream.Dispose()
   at SharpCompress.Readers.AbstractReader`2.Write(Stream writeStream)
   at SharpCompress.Readers.AbstractReader`2.WriteEntryTo(Stream writableStream)
   at SharpCompress.Readers.IReaderExtensions.<>c__DisplayClass4_0.<WriteEntryToFile>b__0(String x, FileMode fm)
   at SharpCompress.Common.ExtractionMethods.WriteEntryToFile(IEntry entry, String destinationFileName, ExtractionOptions options, Action`2 openAndWrite)
   at SharpCompress.Readers.IReaderExtensions.WriteEntryToFile(IReader reader, String destinationFileName, ExtractionOptions options)
   at SharpCompress.Common.ExtractionMethods.WriteEntryToDirectory(IEntry entry, String destinationDirectory, ExtractionOptions options, Action`2 write)
   at SharpCompress.Readers.IReaderExtensions.WriteEntryToDirectory(IReader reader, String destinationDirectory, ExtractionOptions options)
   at ...
```

# Analysis
## Build output
The detailed MSBuild output of building the test project shows:
```
2>  Unified primary reference "System.Buffers, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51".
2>      Using this version instead of original version "4.0.2.0" in "C:\Users\xxx\.nuget\packages\sharpcompress\0.28.1\lib\netstandard2.0\SharpCompress.dll" because AutoUnify is 'true'.
2>      Resolved file path is "C:\Users\xxx\.nuget\packages\system.buffers\4.5.1\ref\net45\System.Buffers.dll".
[...]

PR #621, opened 2021-09-29
Add net461 target to clean up issues with system.* nuget dependencies

inthemedium:
### Reason for change
.NET has a few known issues for libraries that depend on `System.*` and only ship a `netstandard2.0` target

1. .NET has different assembly versions for `netstandard2.0` versus `net461` (see #591). Unless assembly binding redirects are carefully applied, there will be runtime errors.
2. The build output will have tons of "façade" assemblies (see the blogpost in reference 2)

Given the two issues above official guidance is (see reference 1)

> ✔️ CONSIDER adding a target for net461 when you're offering a netstandard2.0 target.
> 
> Using .NET Standard 2.0 from .NET Framework has some issues that were addressed in .NET Framework 4.7.2. You can improve the experience for developers that are still on .NET Framework 4.6.1 - 4.7.1 by offering them a binary that is built for .NET Framework 4.6.1.

So this PR takes the above guidance to make it easier for those of us still stuck on .NET Framework < 4.7.2 :-)

### Example from my recent experience
Due to the recent CVE, I upgraded `SharpCompress` to version  `0.29`. Our code is mostly `.NETFramework`, so we unfortunately have `PackageReferences` to `System.Memory` and `System.Buffers`. Everything appeared to be fine, however at runtime things were failing left and right because most of our projects/nuget dependencies require the latest `System.Memory` and `System.Buffers` while `SharpCompress` requires older versions. This is all avoid requiring a `netstardard2.0` assembly that has an assembly version of `4.0.3.0` (see #591 ) .

After making my change, I found that `.NETFramework` projects work flawlessly without any tricky assembly binding redirects.

### References
1. https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting
2. https://weblog.west-wind.com/posts/2019/Feb/19/Using-NET-Standard-with-Full-Framework-NET

### DIFF
--- a/src/SharpCompress/Common/ArchiveEncoding.cs
+++ b/src/SharpCompress/Common/ArchiveEncoding.cs
@@ -36,12 +36,10 @@ namespace SharpCompress.Common
             Password = password;
         }
 
-#if !NET461
         static ArchiveEncoding()
         {
             Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
         }
-#endif
 
         public string Decode(byte[] bytes)
         {
--- a/src/SharpCompress/Common/ArchiveEncoding.cs
+++ b/src/SharpCompress/Common/ArchiveEncoding.cs
@@ -36,10 +36,12 @@ namespace SharpCompress.Common
             Password = password;
         }
 
+#if !NETFRAMEWORK
         static ArchiveEncoding()
         {
             Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
         }
+#endif
 
         public string Decode(byte[] bytes)
         {
--- a/src/SharpCompress/Algorithms/Alder32.cs
+++ b/src/SharpCompress/Algorithms/Alder32.cs
@@ -4,7 +4,7 @@
 using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-#if !NETSTANDARD2_0 && !NETSTANDARD2_1
+#if !NETSTANDARD2_0 && !NETSTANDARD2_1 && !NETFRAMEWORK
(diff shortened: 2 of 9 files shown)


src/SharpCompress/Common/ArchiveEncoding.cs
```diff
@@ -36,12 +36,10 @@ public ArchiveEncoding(Encoding def, Encoding password)
             Password = password;
         }
 
-#if !NET461
```
inthemedium, line 39: @adamhathcock I removed this `ifdef`. Tests and build still worked fine.

Should I put it back or broaden it to be excluded for all `NETFRAMEWORK`?

adamhathcock: probably `NETFRAMEWORK` as I don't know what's best anymore.

adamhathcock:
Happy to add this back in since someone on legacy framework is testing :)

inthemedium:
That's right! I work on Microsoft Power BI. We'll let you know if something is broken :-)

adamhathcock:
Oops forgot hit review on a comment.  Letting the build run but it all looks good.  I'll get another release out soon

src/SharpCompress/Common/ArchiveEncoding.cs
inthemedium: Ok, I'll put this back

inthemedium: done

adamhathcock:
Released here https://www.nuget.org/packages/SharpCompress/0.30.0

Issue #619, opened 2021-09-26
RarArchive Entry Stream not Copying correctly


CBeatt13:
Having an issue extracting a RAR archive entry to another stream. If I extract the whole archive the files extract correctly. If I attempt to extract a single entry to either a file  or I open the entry stream and then copy that to another stream the file is corrupted. The file I'm extracting is a JPG image, when I attempt to extract the single image the image renders the top 20 lines or so of pixels then the rest of the image is all saved as grey. This only seems to happen sporadically but does happen pretty often for me.

CBeatt13:
Adding some more potentially useful information:

Archive (according to WinRar) is a Rar 4.x file
SharpCompress reads that the entry in question is complete
Archive isn't password protected at all
Have the issue both when reading the first entity immediately, as well as right after using a LINQ statement on all entries to re-order (which I assume is reading all the entities)

adamhathcock:
Without seeing any code, it sounds similar to this problem: https://github.com/adamhathcock/sharpcompress/issues/620

CBeatt13:
I'll get ya a code snippet to show what I'm referring to. Main differences between this and #620 are the following:
I have the issue exclusively with RAR archives
The issue doesn't happen for ALL archives
I don't have to iterate or manipulate the Entry enumerable to see the issue in question
I'm not attempting to iterate the archive's entry enumerable after navigating to the first entry

CBeatt13:
Code I'm using that generates the issue:

'
var archive = ArchiveFactory.Open(archivePath);

foreach (var entry in archive.Entries.Where(e => !e.IsDirectory && e.Key.ToLower().EndsWith(".jpg"))
    .OrderBy(e => e.Key.Contains(Path.DirectorySeparatorChar) ? 0 : 1)
    .ThenBy(e => e.Key))
{
    string fileName = string.Join('_', entry.Key.Split(Path.GetInvalidFileNameChars()));
    imagePath = Path.Combine(Paths.TempDirectory, fileName);
    var tempStream = File.Open(imagePath, FileMode.Create, FileAccess.Write);
    var entryStream = entry.OpenEntryStream();
    entryStream.CopyTo(tempStream);
    entryStream.Close();
    entryStream.Dispose();
    tempStream.Close();
    tempStream.Dispose();

    return true;
}
'

Worth noting I can make it happen still after removing the LINQ query

CBeatt13:
[...]

PR #638, opened 2022-02-15
Rar2 fix with new unit tests that fail on previous build.

Nanook:
I noticed files within WinRar2.8 archives were not decompressing properly. Some were fine, others were fine at the start then corrupt half way through the file. I was able to reproduce this behaviour by using WinRar2.8 to create SharpCompress test archives.

After debugging the streams to confirm the RAR compressed parts were being extracted correctly, I noticed it was the decoding that was failing. I stumbled upon `RarEntry.IsRarV3` property and added `FileHeader.CompressionAlgorithm == 20 || FileHeader.CompressionAlgorithm == 26` to the logic. This forces it to use the UnpackV1 code rather than UnpackV2017. This fixes it.

This PR also includes the UnitTests and Archives. Hopefully it will be useful

I think it fixes this issue also: https://github.com/adamhathcock/sharpcompress/issues/619

Thanks.


adamhathcock (approved):
This looks great.  Thanks for this!

PR #640, opened 2022-02-17
Allow writing a TarHeader directly via TarWriter

### LINKED ISSUE #639: Cannot write Tar Header directly
TarWriter does not allow you to pass a TarHeader structure directly to `TarWriter.Write`, which means that certain POSIX attributes like the `userId` or `groupId` cannot be set

anaisbetts:
This splits TarWriter.Write up and adds a new public method to allow writing TarHeaders directly. This allows users to set POSIX attributes that are otherwise not accessible

Fixes #639


anaisbetts:
Just kidding everything is internal 😭😭😭

adamhathcock:
If you need some internals made public, go for it and I'll review it

Issue #650, opened 2022-03-30
Compression fails in Adler32 on a machine that supports SSE3, but not SSSE3


loop-evgeny:
We're using MongoDB.Driver, which indirectly calls SharpCompress and it fails on a machine with a AMD Opteron(tm) Processor 6176 SE CPU, which supports SSE3, but not SSSE3.

```
System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Runtime.Intrinsics.X86.Ssse3.MultiplyAddAdjacent(Vector128`1 left, Vector128`1 right)
   at SharpCompress.Algorithms.Adler32.CalculateSse(UInt32 adler, ReadOnlySpan`1 buffer)
   at SharpCompress.Algorithms.Adler32.Calculate(UInt32 adler, ReadOnlySpan`1 buffer)
   at SharpCompress.Compressors.Deflate.ZlibCodec.read_buf(Byte[] buf, Int32 start, Int32 size)
   at SharpCompress.Compressors.Deflate.DeflateManager._fillWindow()
   at SharpCompress.Compressors.Deflate.DeflateManager.DeflateSlow(FlushType flush)
   at SharpCompress.Compressors.Deflate.DeflateManager.Deflate(FlushType flush)
   at SharpCompress.Compressors.Deflate.ZlibBaseStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at MongoDB.Bson.IO.ByteBufferStream.EfficientCopyTo(Stream destination)
```

This seems to be because the version of Adler32 used by SharpCompress checks for SSE3 and then uses SSSE3:

```
            if (Sse3.IsSupported && buffer.Length >= MinBufferSize)
            {
                return CalculateSse(adler, buffer); // using SSSE3
            }
```

This is already fixed in the latest version: https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Compression/Zlib/Adler32.cs

That checks `Ssse3.IsSupported` - could you please upgrade to it? That would also fix #645.

loop-evgeny:
Our version of MongoDB driver is using SharpCompress 0.23 which has a [path traversal vulnerability](https://github.com/advisories/GHSA-jp7f-grcv-6mjf) and we cannot upgrade to a patched version due to this issue.

adamhathcock:
I would accept a PR.  Not sure when I'll have time myself

adamhathcock:
[...]

Issue #645, opened 2022-03-07
Adler32 is AGPL-3.0 licensed


tngraf:
The file Adler32.cs contains the statement "Licensed under the GNU Affero General Public License, Version 3".
Due to the copyleft effect of AGPL-3.0 this will put **ALL code** linked dynamically or statically to this file also under AGPL-3.0.
As far as I can see this would affect *ALL* files of SharpCompress.

AGPL-3.0 is also incompatible to MS-PL and Apache-2.0...

Has all this been considered?

adamhathcock:
You are correct.  This needs to be removed as I just read it and it comes from SixLabers/ImageSharp.

It could be updated to the newer version which is Apache 2.0: https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Compression/Zlib/Adler32.cs

adamhathcock:
Fix released in https://www.nuget.org/packages/SharpCompress/0.31.0

PR #651, opened 2022-03-30
Update Adler32 from ImageSharp v2.1.0

loop-evgeny:
Adler32.cs is taken from https://github.com/SixLabors/ImageSharp/commit/09b2cdb83aab0a889dbde8ec2453d605a1be1725 with minimal change to make it build as part of SharpCompress. Fixes https://github.com/adamhathcock/sharpcompress/issues/650 and https://github.com/adamhathcock/sharpcompress/issues/645.


src/SharpCompress/Algorithms/Adler32.cs
```diff
@@ -0,0 +1,420 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+#if !NETSTANDARD2_0 && !NETSTANDARD2_1 && !NETFRAMEWORK
+#define SUPPORTS_RUNTIME_INTRINSICS
```
emz00, line 5: This is my guess of how the #defines should work based on the previous code in Alder32.cs and this in https://github.com/SixLabors/SharedInfrastructure/blob/main/msbuild/props/SixLabors.Global.props

```
    <When Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','netcoreapp3.1'))">
      <!--NETCORE 3.1. NET5.0, and future versions will fallback to this as the closest target.-->
      <PropertyGroup>
...
        <DefineConstants>$(DefineConstants);SUPPORTS_RUNTIME_INTRINSICS</DefineConstants>
        <DefineConstants>$(DefineConstants);SUPPORTS_HOTPATH</DefineConstants>
...
      </PropertyGroup>
    </When>
```

loop-evgeny:
It builds on my machine, the unit tests pass and it fixes #650 for me. Beyond that I can't say whether there any breaking changes.

adamhathcock:
Thank you!

PR #658, opened 2022-04-20
Added Split archive support with unit tests. …

Nanook:
…Added ArchiveFactory.IsArchive() and minor refactor.

ArchiveFactory.cs has new static methods to test if a file is an archive and will also return "out" the ArchiveType. ArchiveFactory.cs was refactored around this to simplify it a bit. A new Open() method was added to allow split archives and multipart rar to be opened - also supporting single file archives too. This method is able to open all types.

A SplitStream class has been added that allows split archives to be read as one file.

Rar4 unit tests have been added, along with split archive tests.

I'm not sure how this change affects some of the other classes and properties with regard to indicating that an archive IsSplit etc. A minor enhancement would be be able to give the library the first part of a multipart or split archive and have the library find the other parts. Currently you have to pass them all in to the new Open() method.

Hope this is useful for the repo.


adamhathcock:
Wow, blown away by this.  Thank you!

PR #661, opened 2022-04-25
Added multipart Zip support (z01...). Added IEntry.IsSolid

Nanook:
**Multi Volume Zip support for zip and zipx**
Unit tests added for multipart archives created with WinZip and InfoZip. They are uncompressed due to the minimum part size being 65k.

The implementation is a little bit clunky. It requires the parts to be passed in in the order zip,z01 / zipx,zx01 ..., the zip is actually the last part of the archive. Once the central directory has been located and the archive is being opened the zip is switched to the end of the volumes for processing. Is seems more logical the user would select the zip/zipx not the z01/zx01. Feel free to reimplement this as your knowledge of zip is better than mine ;-)

**IEntry.IsSolid for rar and 7zip**
It's useful to know which files in an archive can be seeked and extracted. The Rar format sets the first file to solid=false and subsequent files in a solid set to solid=true. This allows solid-false files to be opened with seek. The 7zip implementation follows this behaviour. IsSolid already existed for Rar Entries, but it was never populated when using Archive.Open.

Cheers ;-)


.gitignore
```diff
@@ -18,3 +18,4 @@ tools
 
 .DS_Store
 *.snupkg
+/tests/TestArchives/6d23a38c-f064-4ef1-ad89-b942396f53b9/Scratch
```
adamhathcock, line 21: accident?

src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
```diff
@@ -101,11 +101,23 @@ protected override IEnumerable<SevenZipArchiveEntry> LoadEntries(IEnumerable<Sev
         {
             var stream = volumes.Single().Stream;
             LoadFactory(stream);
+            var entries = new SevenZipArchiveEntry[database._files.Count];
             for (int i = 0; i < database._files.Count; i++)
             {
                 var file = database._files[i];
-                yield return new SevenZipArchiveEntry(this, new SevenZipFilePart(stream, database, i, file, ReaderOptions.ArchiveEncoding));
+                entries[i] = new SevenZipArchiveEntry(this, new SevenZipFilePart(stream, database, i, file, ReaderOptions.ArchiveEncoding));
```
adamhathcock, line 108: I like my yield returns for streaming but 7Zip always requires random access so no big deal

src/SharpCompress/Archives/Zip/ZipArchive.cs
```diff
@@ -131,33 +184,41 @@ internal ZipArchive(Stream stream, ReaderOptions readerOptions)
...
+            foreach (var s in st)
+                yield return new ZipVolume(s, base.ReaderOptions);
         }
 
         protected override IEnumerable<ZipArchiveEntry> LoadEntries(IEnumerable<ZipVolume> volumes)
         {
-            var volume = volumes.Single();
-            Stream stream = volume.Stream;
-            foreach (ZipHeader h in headerFactory.ReadSeekableHeader(stream))
+            var vols = volumes.ToArray();
```
adamhathcock, line 198: More eager loading here.  I guess we're saying multi-volume has to have access to all streams/volumes at the start?  RAR doesn't need to operate this way but I guess Zip does because you have to have access to the directory at the end?

adamhathcock (commented):
The big thing I'm mulling over is how to be more explicit about the access/streams requirements for multi-zip.

Multi-RAR only needs sequential access so can stream in the streams.  Zip has to have the directory always?  7Zip has to have random access to everything.

src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
Nanook: I just realised

`                foreach (var entry in group)
                {
                    entry.IsSolid = isSolid;
                    isSolid = true; //mark others in this group as solid - same as rar behaviour.
                    yield return entry;
                }
`
Another option is to allow the SevenZipArchiveEntry to have access to the entries collection so IsSolid can calculate it when called. Or add an IsSolid property to CFileItem. I'm happy to rework it.

src/SharpCompress/Archives/Zip/ZipArchive.cs
Nanook: It needs the first and last item at a minimum. the z01 contains the header, the zip (last part) contains the central directory. Unlike rar there's no header per file. Files that span volumes just continue in to the next file. Each entry has the diskpart number where the file starts.

You could open the next part when it's required like with Rar. This PR could be classed as a rough first effort to demonstrate an understanding of the format, I don't have a good enough understanding of all that the library is capable of to make (or attempt) significant change. It would be good to cater for multi volume and split in one set of logic behind an interface where parts can be loaded as required etc.

.gitignore
Nanook: Yes, good catch.

src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
adamhathcock: maybe being a big more clever is warranted but 7Zip is my most hated format so I'm ambivalent.

src/SharpCompress/Archives/Zip/ZipArchive.cs
adamhathcock: yeah, I think there's more thought needed but feels like this can be a good first pass

Nanook:
@adamhathcock. Thanks for merging this. I have a new version to push soon with nicely integrated split and multi volume support. It's able to look the files up on load and is more fitting with the model. I'll PR it over the weekend.

PR #662, opened 2022-04-30
Properly integrated zip multivolume and general split support.

Nanook:
Proper multi volume zip and generic split support.

SplitStream has been renamed to SourceStream and supports multi and split archives. It's able to lookup missing parts.

- The 5 Archive types have all been changed to add more static Open() methods.
- The 5 Archive types now only have one constuctor for reading that accepts a SourceStream. Writing are left unchanged
- The 5 Archive types now only have one LoadVolumes method that accepts a SourceStream

A new static method has been added to ArchiveFactory. It's a utility method called GetFileParts() that will take a file name/FileInfo and return all the files that are part of the set. It might be handy for people wanting to scan folders for archives and gather the other part names. 

ArchiveFactory is cleaner as the multipart / split code that was in there is no longer required.

All tests are passing with no changes. More have been added to demonstrate new behaviour.

Cheers ;-)


adamhathcock (approved):
Looks great.  Looks like I should do a release

Issue #670, opened 2022-06-12
[proposal] IArchiveFactory


vpenades:
Right now, most archives are wrapped around interfaces, and it's useful to use these interfaces to wrap around custom or third party archive formats.

But right now the archive formats supported by SharpCompress is limited by ArchiveFactory static class

I believe it could be useful to have an `IArchiveFactory `interface and a `RegisterFactory(IArchiveFactory customArchive)` method so developers could inject support for additional archive formats.

adamhathcock:
That's not a bad idea though I'd like people to contribute missing formats if possible!

vpenades:
I think you're covering most formats already?

There's some legacy formats that might be included, like ARJ, ACE that I think nobody uses anymore, but would it make sense to include them in the core library?  In fact in these cases it would make more sense to include them as a separated library called SharpCompress.Legacy  or something like that -this is just an example-

Then, I think people could use this extension to give support for game archives, like WAD, but I don't think you want to integrate these into the main library either.

Anyway, I would like to have this extensibility feature to support a proprietary archive we use for scientific data analysis; it's usually thousands of anotated small files which is convenient to move around as a single archive.

So, if you agree, I could contribute a PR with this interface

adamhathcock:
Yeah, sure...make a PR to have a proper registry of formats.  I may refactor more after seeing your PR 🤔 

I don't have a big problem with legacy formats in general though.

vpenades:
Since it's been merged I think I can close this issue, unless there's something else that might be interesting to add to the interface...

PR #671, opened 2022-06-13
Introduced IArchiveFactory

vpenades:
As agreed in #670 , I've introduced the `IArchiveFactory` interface, with the bare minimum methods required to keep all the funcionality, and the factory implementations for each archive type.

Also, added the `RegisterFactory` method in `ArchiveFactory` and adjusted the `Open` method to use the registered factories instead of the fixed list.

remaining issues:

Not sure to which factory should I assign some extensions lik .LZ

In ArchiveFactory there's some more Open methods that still use the fixed list, I guess they should be redirected to the Open method that uses the factory collection.


vpenades:
I think the interface is pretty much ready for review

adamhathcock (approved):

PR #672, opened 2022-06-15
Corrected skip-marker on skip of uncompressed ZIP file with missing size informations.

### LINKED ISSUE #477: Failed to extract onedrive zipped file
[Samplecode.txt](https://github.com/adamhathcock/sharpcompress/files/3587678/Samplecode.txt)
I have a zip file zipped by OneDrive. When i call the ExtractAllFiles method and i iterate through the files (the zip has 4 files) a EndOfStream Exception is thrown after the first entry.

I attached the File and some sampleCode to Test it with NUnit
[OneDrive.zip](https://github.com/adamhathcock/sharpcompress/files/3587680/OneDrive.zip)

MartinDemberger:
I did a first try to fix this issue. I'm not sure if my fix is correct. This fixes #477 

My test is not working now because there is a `EndOfStreamException` thrown which I don't know how to avoid or if this is correct.


Erior:
The Zip file in your test exposes two problems with the library
1. The archive.ExtractAllEntries does not propagate the compressed/uncompressed sizes of entries
2. Your test file is generated in streamed mode, the size information is missing (set to 0), trying to skip will fail in current implementation. (need the size from the above)

One way to solve this would be to handle the searching for the Data descriptor, this can be tricky, you could embedd a zip in a zip and you might detect the wrong one.

This would be an example implementation
[https://github.com/Erior/sharpcompress/tree/feature/Failed-to-extract-onedrive-zipped-file](url)

MartinDemberger:
@Erior thanks for your hint. This l guided me in the right direction.

The problem was a unset skip marker so the inner file was skipped several files which lead to the exception.

adamhathcock (approved):

adamhathcock:
thanks for the fix...it looks sound

PR #674, opened 2022-06-19
Suppress nested NonDisposingStream

MartinDemberger:
Suppresses the creation of a `NonDisposingStream` which wrappes another `NonDisposingStream` with same settings


adamhathcock:
Thanks!

Issue #636, opened 2022-02-11
Reading from 7z entry stream using `Read(Span<byte>)` might read too many bytes


Molinarius:
With .NET 5 and .NET 6, I noticed that reading from an entry stream from certain 7z file reads too many bytes when the `Read(Span<byte>)` overload is used.

E.g.:

    using (var archive = SevenZipArchive.Open("test.7z"))
    {
        var entry = archive.Entries.First(entry => entry.Key == "test.txt");
        using (var stream = entry.OpenEntryStream())
        {
            byte[] buffer = new byte[1024];
            int bytesRead = stream.Read(buffer.AsSpan()); // stream.Read(buffer, 0, 1024) works
            Console.WriteLine(bytesRead);
        }
    }

With the sample 7z file attached (I had to put it into a zip file [test.zip](https://github.com/adamhathcock/sharpcompress/files/8047989/test.zip) to be able to upload it), this reads 18 bytes instead of the expected 12 bytes. The `Read(byte[], int, int)` overload works for me as expected.

Using version 0.30.1 from NuGet.

PR #675, opened 2022-06-21
ReadOnlySubStream overrides and adds logic #636

Erior:
ReadOnlySubStream overrides and adds logic to Read byte[], it needs to have same logic for Span<byte> for consistency.
#636


adamhathcock:
Thanks!

PR #680, opened 2022-07-01
Error extracting uncompressed zip with ExtractAllEntries

MartinDemberger:
There is an error when uncompressing a uncompressed ZIP file via ExtractAllEntries.

Currently there is only a test that reproduces the error.


adamhathcock:
Thanks for making the breaking test!

Hopefully someone will get a chance to look at this soon.

Erior:
Feels like I'm using the github tool wrong, could I have built upon MartinDemberger's pull here somehow instead of created #727 as a pull/fix?

Issue #677, opened 2022-06-22
RAR archives - Cannot get volume number


lchegde:
Is it possible to read the internal volume number for each file in a multi volume archive?
I cannot find this information for RarArchive presently, is this information available using sharpCompress?

If not, is this something that could be supported?

Example from `rar.exe ls`, showing the volume number in question for a couple of volume files in the same archive:

```
Archive: <rar-file first volume>
Details: RAR 5, volume 1

 Attributes      Size     Date    Time   Name
----------- ---------  ---------- -----  ----
 -rw-r--r--   1161022  2022-04-24 07:12  filename.jpg
 -rw-r----- 1641030031  2022-04-24 07:10  filename.mp4
----------- ---------  ---------- -----  ----

Archive: <rar-file second volume>
Details: RAR 5, volume 2

 Attributes      Size     Date    Time   Name
----------- ---------  ---------- -----  ----
 -rw-r----- 1641030031  2022-04-24 07:10  filename.mp4
----------- ---------  ---------- -----  ----

````

adamhathcock:
it's possible but it's not that simple.  An entry can be spread across multiple volumes.  The data is in the FilePart classes and could be exposed somehow: https://github.com/adamhathcock/sharpcompress/blob/master/src/SharpCompress/Common/Rar/RarFilePart.cs

Nanook:
https://github.com/adamhathcock/sharpcompress/blob/master/src/SharpCompress/Common/IEntry.cs

Now has the first and last volume index. This index can be used to get the files filenames using archive.Volumes[index].FileName

Issue #643, opened 2022-02-26
IEntry.Crc returns a negative number


roland5572:
IEntry.Crc Property returns a negative number (eg: -1626075223). 

When loading some .gz files, the CRC property will return a negative number. I'm not sure if negative numbers are currently the best solution for sharpcompress. But while converting this CRC to a string, an error occurs.

Test Code:
C# / .net 6.0 / Visual Studio 2022

```
using SharpCompress.Archives;

var filePath = @"D:\001\001.gz";

using var stream = File.OpenRead(filePath);
var archive = ArchiveFactory.Open(stream);

foreach (var p in archive.Entries)
{
    var data = p.Crc;

    var text = data.ToString("X");
    Console.WriteLine(text);
}
```

Test Result:
```
FFFFFFFFD8BF5DCD
```

String conversion has extra "FFFFFFFF" when CRC is **negative number**.
There is a detailed explanation about this part on stackoverflow.

https://stackoverflow.com/questions/32940417/unsigned-crc-32-for-python-to-match-javas-crc-32

Do you think it is necessary to normalize all CRCs to **positive integers**?

Nanook:
I've fixed this and will submit a fix.

roland5572:
@Nanook Thank you very much.

Issue #632, opened 2022-01-23
Cannot get RAR version from the library API


roland5572:
Currently encrypted RAR 5.0 files cannot be read due to sharpcompress. But currently sharpcompress currently supports the reading of encrypted files of 4.0.

If the file version number of RAR can be directly read in the program, then the end user can be directly prompted that the encrypted file of 5.0 is not supported. At present, since the version number cannot be obtained, the end user can never judge which files are in the process of using can be read normally.

If a Property API can be provided to display the version number of RAR or other formats, after RAR 6.0 or 7.0 comes out in the future, programmers can use this Property to prompt users which files can be read normally.

adamhathcock:
Could be a good idea to add

Nanook:
After some investigation it appears there isn't a 100% way to determine which version of WinRar/Rar created an archive. I've implemented a solution which I will be submitting that does it based on the compression algorithm used on the files. Two properties have been added to the RarArchive class:
- MinVersion
- MaxVersion

Some examples:
- If the algorithm is 20 then the archive was created by MinVersion=2 MaxVersion=2
- If the algorithm is 29 then the archive was created by MinVersion=3 MaxVersion=4
- If the algorithm is 50 then the archive was created by MinVersion=5 MaxVersion=6

Does this seem fair?

PR #682, opened 2022-07-15
RarArchive has Min/MaxVersion. RarEntry has Volumne Indexes. GZ CRC fix.

Nanook:
RAR archives - Cannot get volume number #677

Cannot get RAR version from the library API #632

IEntry.Crc returns a negative number #643

Sorry it's 3 in 1. I can separate if required.


Nanook:
These changes are quite small. Adding the index to the volume touched a lot of classes. There should be no breaking changes though.

The GZ negative CRC fix was just a case of changing the type from int to uint. It's a exposed long as public. I took the liberty to fix the length too after checking the Gzip spec.

Unit tests have been added to demonstrate functionality with existing test archives.

adamhathcock:
Thanks!

I was hoping the rar version could be an enumeration or something but if it can't be then this looks good.

Nanook:
It can be, but it needs a little thought (though I might be over thinking it.

Currently algo 50 is WinRar5To6 for example. If later it became WinRar5To7 would that be considered a breaking change? Or I could go the Enum Flags route and just mask WinRar5 | WinRar6 | WinRar7.

What is you view? :)

adamhathcock:
Hmm maybe it's more complex than I realize.  Can add/change later.

PR #683, opened 2022-07-18
WriteAll: use delegate instead of Expression

OwnageIsMagic:


adamhathcock (approved):

src/SharpCompress/Writers/IWriterExtensions.cs
```diff
@@ -37,7 +37,7 @@ public static void WriteAll(this IWriter writer, string directory, string search
         public static void WriteAll(this IWriter writer,
                                     string directory,
                                     string searchPattern = "*",
-                                    Expression<Func<string, bool>>? fileSearchFunc = null,
+                                    Func<string, bool>? fileSearchFunc = null,
```
OwnageIsMagic, line 40:
```suggestion
                                    Predicate<string>? fileSearchFunc = null,
```

adamhathcock (changes requested):
Looks like func is required unless you change the `Where`

OwnageIsMagic (commented):
WriteAll: revert 109a7c12ead9893

adamhathcock (approved):

PR #684, opened 2022-07-18
Include license in nuget package

daverant:
Some licensing tools I use (Snyk) were failing to identify SharpCompress' MIT license as it's not bundled into the Nuget package. This bundles LICENSE.txt in the root of the repo into the package following guidance [here](https://docs.microsoft.com/en-gb/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file).

Before:
image

After:
image

Could alternatively include an SPDX license expression but this [wouldn't include specific wording for the author's copyright](https://licenses.nuget.org/MIT)
```
<PropertyGroup>
    <PackageLicenseExpression>MIT</PackageLicenseExpression>
</PropertyGroup>
```

### DIFF
--- a/src/SharpCompress/SharpCompress.csproj
+++ b/src/SharpCompress/SharpCompress.csproj
@@ -31,7 +31,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <None Include="..\..\LICENSE.txt" Pack="true" PackagePath=""/>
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):
Thanks...I had tried to do this myself but could not find the correct combination of options :(

PR #686, opened 2022-07-28
Mitigation of problems

Erior:
Mitigations of the problems found, we still need to either make the Archive tell Reader about the CompressedSizes of the archive or make a search for DataDescriptor, or both since that would increase performace in general, no need to decompress if we know the compressed size, current code is already handling that.


adamhathcock (approved):

adamhathcock:
thanks!

PR #687, opened 2022-07-30
SourceStream Position counting bug fix

Erior:
Encountered a position counting bug, it add the stream size even if the stream is not switched.
This should fix it and I also removed the workaround I had.


adamhathcock (approved):

adamhathcock:
Thanks!

PR #696, opened 2022-09-15
Ignores UnicodePathExtra if forced encoding is specified

stakira:
- It may be surprising, but probably due to the history of Windows, it is not unusual to encounter zip files compressed with wrong unicode path info.
- My program provides a UI for users to preview decoded paths and choose a proper encoding before extracting. However, `UnicodePathExtra` always overriding specified encoding makes this impossible for some files.
- `Forced` encoding doesn't seem to be used anywhere, other than shadowing `Default`. That basically makes the two fields equivalent. Shouldn't it override `UnicodePathExtra`? Anyway, I think this is the intended behavior, instead of the current one.


adamhathcock:
Seems valid to me.  Can't say how all these cases will affect things

PR #697, opened 2022-09-29
Added support for reading comment header for Rar v5 archives

IngBertolini:
Hello! Like for the Zip archives i tried to add the Comment property in RarVolume, in order to read the comment header from the archive. It only works with Rar v5 archives though


adamhathcock (approved):
Looks great and to spec!

Had to look it up as the CMT part was weird but that's what it is.

PR #706, opened 2022-11-27
Use PackageLicenseExpression instead of PackageLicenseFile

andreas-eriksson:
This makes it easier to validate the license used as it can be automatically checked by tools.


adamhathcock (approved):
Thanks!

PR #707, opened 2022-11-28
Added SupportsRandomAccess property to IArchiveFactory

vpenades:
I've been using tgz files and I've realized they require being opened using the IReader interface to read them sequentially.

I thought it could be useful to embed that knowledge in the IArchiveFactory.

So both TAR and GZip factories have that property set to false.

This way, a program that wants to open any kind of archive, would know beforehand which is the preferred way of reading the archive and adjust accordingly.

Also added the "tgz" extension to tar factory....  I guess it would also require "tar.gz"  but I'm hesitant to add extensions with extra dots.

### DIFF
--- a/src/SharpCompress/Archives/GZip/GZipArchiveFactory.cs
+++ b/src/SharpCompress/Archives/GZip/GZipArchiveFactory.cs
@@ -13,6 +13,9 @@ namespace SharpCompress.Archives.GZip
         public string Name => "GZip";
 
         /// <inheritdoc/>
+        public bool SupportsRandomAccess => false;
+
+        /// <inheritdoc/>
         public IEnumerable<string> GetSupportedExtensions()
         {
             yield return "gz";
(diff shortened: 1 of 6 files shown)


vpenades:
Hmm... I think there's a better solution for this... closing!

adamhathcock:
I like the idea of trying to tell people when using the API.  Keep going!

vpenades:
I've noticed you really don't know if you need to read sequentially or you can access randomly until you have opened the archive... correct me if I'm wrong:

Plain .TAR files can actually be read randomly....  but tar archives inside a gzip archive can't

So, if that's the case, the "SupportsRandomAccess" property should go into the IArchive, not in the IArchiveFactory interface, right?

Also, wouldn't the IsSolid property would be enough for this?  (in that case, IsSolid should be true for .tar.gz files)

If IsSolid serves a different purpose, then, IArchive definitely needs a SupportsRandomAccess

Any thoughts?

PR #725, opened 2023-02-18
Feature/7z delta

Erior:
After some conversation the delta filter functionality might need an implementation, refer to #526


Erior:
Wrong branch

PR #727, opened 2023-02-19
Implement Searching Data Descriptor stream issue/pull #680

Erior:
A solution for the problem described in #680


adamhathcock:
Can you resolve this? I've been away for personal reasons.

Erior:
The xz pull introduced a potential span position counting bug, did not add checks for it, but merged and fixed it. The code for dealing with the non compressed streamed zip files still works okay in the tests.

adamhathcock (approved):
Not sure I like it but it works well as far as I can tell 👍

PR #731, opened 2023-03-08
LHA archive decompression support added

TwanVanDongen:
LHA Archive support added for decompression of methods lh0, lh5, lh6 and lh7. Others will be added when test-archives are created. Implementation taken from [LHA Decompressor for .NET](https://github.com/benbaker76/LHADecompressor).

### DIFF
--- a/src/SharpCompress/SharpCompress.csproj
+++ b/src/SharpCompress/SharpCompress.csproj
@@ -2,9 +2,9 @@
 	<PropertyGroup>
 		<AssemblyTitle>SharpCompress - Pure C# Decompression/Compression</AssemblyTitle>
 		<NeutralLanguage>en-US</NeutralLanguage>
-		<VersionPrefix>0.35.0</VersionPrefix>
-		<AssemblyVersion>0.35.0</AssemblyVersion>
-		<FileVersion>0.35.0</FileVersion>
+		<VersionPrefix>0.36.0</VersionPrefix>
+		<AssemblyVersion>0.36.0</AssemblyVersion>
+		<FileVersion>0.36.0</FileVersion>
 		<Authors>Adam Hathcock</Authors>
 		<TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net7.0;net8.0</TargetFrameworks>
 		<AssemblyName>SharpCompress</AssemblyName>


.gitignore
```diff
@@ -19,3 +19,6 @@ tools
 .DS_Store
 *.snupkg
 /tests/TestArchives/6d23a38c-f064-4ef1-ad89-b942396f53b9/Scratch
+/src/SharpCompress/Readers/ReaderFactory.cs
```
adamhathcock: This doesn't seem right

src/SharpCompress/Archives/Lha/LhaArchive.cs
```diff
@@ -0,0 +1,115 @@
+#nullable disable
```
adamhathcock: Not a fan but can overlook for this PR

tests/SharpCompress.Test/SharpCompress.Test.csproj
```diff
@@ -4,6 +4,9 @@
     <AssemblyName>SharpCompress.Test</AssemblyName>
     <PackageId>SharpCompress.Test</PackageId>
   </PropertyGroup>
+  <ItemGroup>
+    <Compile Remove="Lha\LhaReaderTests.cs" />
```
adamhathcock: Seems wrong

adamhathcock (changes requested):
Never heard of LHA!

Some minor things but the major is the LICENSE which is LGPL and we're MIT

Unless you got permission, I can't accept code from https://github.com/benbaker76/LHADecompressor/tree/master unfortunately

.gitignore
TwanVanDongen: yep, accidentally removed it, but it's back now!

src/SharpCompress/Archives/Lha/LhaArchive.cs
TwanVanDongen: also not too happy about it, but saw that it was accepted elsewhere; certainly want to fix it where code is accepted

tests/SharpCompress.Test/SharpCompress.Test.csproj
TwanVanDongen: Started implementing also the reader function, but removed it in order to get the 'Archive' support in first.

TwanVanDongen:
> Never heard of LHA!
> 
> Some minor things but the major is the LICENSE which is LGPL and we're MIT
> 
> Unless you got permission, I can't accept code from https://github.com/benbaker76/LHADecompressor/tree/master unfortunately

Received confirmation from Headkaze's to use his C# port of LHADecompressor (" I have no problem with you using my port. I believe it's the same license as the original source. If it complies then it shouldn't be a problem."). This however only resolved the porting, now trying to connect the original Java developer (Nobuyasu SUEHIRO)  with the same question.

vpenades:
Suggestion:  If the problem is only the license, an alternative approach would be this:

Create a separate project library called **SharpCompress.ArchiveFormat.LHA** (or something like that) which would use the LGPL license, so it would be license compatible.

Users willing to use LHA would require:

- To be also license compatible with LGPL
- Reference SharpCompress
- Reference SharpCompress.ArchiveFormat.LHA
- register the LHA factory at application startup

adamhathcock:
that's a good call, to do that.  Can register factories at start up now with little code

Issue #736, opened 2023-03-24
TarHeaderFactory Infinite loop


Ilanlido:
A possible Infinite loop can occur,
Let's look at https://github.com/adamhathcock/sharpcompress/blame/master/src/SharpCompress/Common/Tar/TarHeaderFactory.cs#L51
When the stream that is being used has been closed - for example if the connection was cut, it will keep throw an IOException inside https://github.com/adamhathcock/sharpcompress/blame/master/src/SharpCompress/Common/Tar/TarHeaderFactory.cs#L24
And then and infinite loop of nulls will be here 
https://github.com/adamhathcock/sharpcompress/blob/master/src/SharpCompress/Common/Tar/TarEntry.cs#L62

adamhathcock:
PRs are welcome

svenclaesson:
[zoneinfo-2010g.zip](https://github.com/adamhathcock/sharpcompress/files/11418820/zoneinfo-2010g.zip)

This file contains a corrupt tar file that also triggers this infinite loop issue.

PR #737, opened 2023-03-28
Check for broken file #736

Erior:
Guess this would be one way to handle #736


Erior:
Not sure why this change would make unpack_hpp.cs fail for rar

adamhathcock (approved):

PR #740, opened 2023-05-10
Make ArchiveFactory.IsArchive(Stream, ...) public. Fix #739

### LINKED ISSUE #739: SharpCompress.Archives.ArchiveFactory.IsArchive(Stream, out ArchiveType?) is private
## Issue description

We sometimes need to know whether a particular file is an archive of some kind or not an archive. If it is a plain file on the filesystem, SharpCompress provides `SharpCompress.Archives.ArchiveFactory.IsArchive(string, out ArchiveType?)` for the purpose. If it is _not_ a file on the filesystem, e.g., it is an in-memory data or a _part_ of a file on the filesystem embedded in a larger file format, the `IsArchive(string, out ArchiveType?)` doesn't work, unless we first write the data into a single temporary file.

The `SharpCompress.Archives.ArchiveFactory` class has another method `IsArchive(Stream, out ArchiveType?)`, whch just fits the purpose. However, the `IsArchive(Stream, out ArchiveType?)` method is private and not available for user programs.

I consider it is a bug (or a design flaw.)

## Proposed solution

Make `SharpCompress.Archives.ArchiveFactory.IsArchive(Stream, out ArchiveType?)` public.

AlissaSabre:
This is the straightforward _fix_ to solve the issue #739 (which I raised).

Thank you.

### DIFF
--- a/src/SharpCompress/Archives/ArchiveFactory.cs
+++ b/src/SharpCompress/Archives/ArchiveFactory.cs
@@ -175,7 +175,7 @@ public static class ArchiveFactory
         return IsArchive(s, out type);
     }
 
-    private static bool IsArchive(Stream stream, out ArchiveType? type)
+    public static bool IsArchive(Stream stream, out ArchiveType? type)
     {
         type = null;
         stream.CheckNotNull(nameof(stream));


adamhathcock:
Thank you!

PR #746, opened 2023-06-09
Added simple example

rodesfl:
Had some difficult figuring this one out, as this is so straight forward and is the simplest use case, maybe should be included at the top of the docs ?


adamhathcock (approved):

adamhathcock:
thank you.  What samples you'd like to add would be great

Issue #728, opened 2023-02-20
When decompressing a file in 7z, there will be a problem of performance degradation


roland5572:
I have conducted some tests, and sharpcompress does suffer from **performance degradation** when reading some 7z files. 

For example, there are 100 files in a 7z archive. When it reads the previous files, such as the first 10, the speed is very fast, but the more it reads the latter files, the slower the decompression speed will be. 

There seems to be a rule that the further the file is in the back, the slower the decompression will be.

For example, **if you have a 300MB 7z file**, you extract one of the 100 files, and it appears in the 10th file order, you will read it very quickly, but if you read the **70th** file, then Even if they have the same file size, the decompression speed will become very slow.

adamhathcock:
PRs are welcome.  I've been away for personal reasons.

Nanook:
The speed degradation described above could be due to 7zip grouping files with "Solid Block Size" to achieve better compression.  Files at the end of these blocks cannot be directly decompressed. Earlier files must be decompressed first.

roland5572:
> The speed degradation described above could be due to 7zip grouping files with "Solid Block Size" to achieve better compression. Files at the end of these blocks cannot be directly decompressed. Earlier files must be decompressed first.

@Nanook Thank you, I use WinRAR to test the same files, and WinRAR works normally, it does not have this bug.

Nanook:
It's not necessarily a bug, but rather a trade off between compression and flexibility. Rar also has SOLID mode which makes the full archive non seekable. 7zip's is at least configurable. Just for info :)

roland5572:
@Nanook No, I meant to use WinRAR to decompress **.7z** files. It has almost no problem of slowing down. WinRAR is fast at unpacking both .zip and .7z files.

Nanook:
Thanks for the info. Perhaps the SharpCompress implementation can be improved. I'll take a look next time I'm in there.

roland5572:
Thank you very much, looks like this is an old bug, detailed data in this post.

post url:
https://github.com/adamhathcock/sharpcompress/issues/399

Erior:
Looking at the cases it seems to be using archive.ExtractAllEntries() vs archive.Entries.Where(entry => !entry.IsDirectory)

Performance might not be great but the latter does a skip to find the entry you try to decompress..the more files in the 7z file the more to redo/decompress to get to your file.

archive.ExtractAllEntries()
while( reader.MoveToNextEntry()) )
....

[...]

PR #750, opened 2023-07-19
Add fast `ExtractToDirectoryAsync` extension method on `IArchive`

FlsZen:
As mentioned in #728 (and I believe some other issues), the `IArchive.WriteToDirectory(...)` extension method can be very slow for `.7z` files. This extension utilized `IArchive.ExtractAllEntries()` and `IReader` so is fast for .7z files.

Being async has nothing to do with the performance, rather, it's because I use this extension method from an async method and need it to perform the long-running work asynchronously. Ideally `IArchive.ExtractAllEntries()` and `IReader.WriteEntryTo(Stream)` would be async as well and then the whole Task.Run wouldn't be needed and then it would make more sense to be async.

It currently doesn't support `ExtractionOptions` or reporting progress through `IArchive`'s events (although it does report progress through a callback). But, it works great as-is for my use case, which I suspect is pretty common, and I wanted to share.

### DIFF
--- a/src/SharpCompress/Archives/IArchiveExtensions.cs
+++ b/src/SharpCompress/Archives/IArchiveExtensions.cs
@@ -60,8 +60,7 @@ public static class IArchiveExtensions
 
             // Create each directory
             var path = Path.Combine(destination, entry.Key);
-            if (Path.GetDirectoryName(path) is { } directory
-                && seenDirectories.Add(path))
+            if (Path.GetDirectoryName(path) is { } directory && seenDirectories.Add(path))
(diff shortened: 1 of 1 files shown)


adamhathcock (changes requested):
`Task.Run` isn't for this library

FlsZen:
I've removed the Task aspect in the implementation.

It still feels a bit odd to me since it has its own progress reporting and doesn't fire the progress events on `IArchive`.

adamhathcock (approved):

Issue #751, opened 2023-07-21
.tar.7z support


SupinePandora43:
stacktrace:
```
SharpCompress.Compressors.LZMA.DataErrorException: Data Error
   at SharpCompress.Compressors.LZMA.LzmaStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at SharpCompress.Compressors.LZMA.Utilites.Utils.ReadExact(Stream stream, Byte[] buffer, Int32 offset, Int32 length)
   at SharpCompress.Common.SevenZip.ArchiveReader.ReadAndDecodePackedStreams(Int64 baseOffset, IPasswordProvider pass)
   at SharpCompress.Common.SevenZip.ArchiveReader.ReadDatabase(IPasswordProvider pass)
   at SharpCompress.Archives.SevenZip.SevenZipArchive.LoadFactory(Stream stream)
   at SharpCompress.Archives.SevenZip.SevenZipArchive.LoadEntries(IEnumerable`1 volumes)
   at SharpCompress.Archives.AbstractArchive`2..ctor(ArchiveType type, SourceStream srcStream)
   at SharpCompress.Archives.SevenZip.SevenZipArchive..ctor(SourceStream srcStream)
   at SharpCompress.Archives.SevenZip.SevenZipArchive.Open(FileInfo fileInfo, ReaderOptions readerOptions)
   at SharpCompress.Archives.SevenZip.SevenZipArchive.Open(String filePath, ReaderOptions readerOptions)
```
file in question: [ultralightcore-bin-41a04e01-linux-x64.7z](https://ultralightcore-bin.sfo2.cdn.digitaloceanspaces.com/ultralightcore-bin-41a04e01-linux-x64.7z)
file sha256: `c6dd7e4ba5867fe763109263a5af8da40cf3e5cf46b91e4892eac76a49ec0746`

code to replicate:
```cs
SevenZipArchive.Open(file);
```

~~for some reason, i can't open it in `Ark`, but can in `Dolphin` 😕~~
Turns out it's `.tar.7z` archive!

adamhathcock:
should easy to add to detection alongside gzip and bzip2

btomblinson:
@adamhathcock any progress on this?

btomblinson:
@adamhathcock this can be closed or do we want to add more features before we say we fully support it?

btomblinson:
Pull request #763 closes this

PR #756, opened 2023-08-26
Add .tar.7z support

btomblinson:
Candidate for #751

I used Ubuntu subsystem on Windows to create a .tar.7z file, and was able to open and read it via the test


src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
```diff
@@ -105,7 +105,7 @@ public static SevenZipArchive Open(Stream stream, ReaderOptions readerOptions =
     /// <param name="srcStream"></param>
     /// <param name="options"></param>
     internal SevenZipArchive(SourceStream srcStream)
-        : base(ArchiveType.SevenZip, srcStream) { }
+        : base(ArchiveType.Tar, srcStream) { }
```
adamhathcock, line 108: This isn't right.  We can't change archive types.

What do tars for gz and bz2 do?

adamhathcock (changes requested):

src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
btomblinson: The GZipArchive does the same thing, that was the only way that the test was able to succeed for both GZip and SevenZip

btomblinson:
```
    /// <summary>
    /// Constructor with a SourceStream able to handle FileInfo and Streams.
    /// </summary>
    /// <param name="srcStream"></param>
    /// <param name="options"></param>
    internal GZipArchive(SourceStream srcStream)
        : base(ArchiveType.Tar, srcStream) { }

```

PR #763, opened 2023-09-17
#751 Add .tar.7z support

btomblinson:
@adamhathcock I started over after being unable to merge the latest changes to my fork, got rid of the ArchiveType change as well so should be good to go


adamhathcock:
Thanks!

Issue #761, opened 2023-09-14
Checking if a 7z file is encrypted will crash


njlr:
Create a `7z` file:

```bash
$ 7zz a test.7z
```

Check the entries are encrypted:

`Scratch.fsx`

```fsharp
#r "nuget: SharpCompress, 0.33.0"

open SharpCompress
open SharpCompress.Archives

let archive = ArchiveFactory.Open("test.7z")

for e in archive.Entries do
  printfn $"IsEncrypted %b{e.IsEncrypted}"
```

```bash
$ dotnet fsi ./Scratch.fsx
```

It will crash:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at SharpCompress.Common.SevenZip.SevenZipFilePart.get_IsEncrypted()
   at SharpCompress.Common.SevenZip.SevenZipEntry.get_IsEncrypted()
```

Erior:
To see if I recreated the issue correctly, if you check if it is a directory before checking the IsEncrypted flag
  `if not e.IsDirectory then`
 Does that do anything for you?

njlr:
This does not crash:

```fsharp
#r "nuget: SharpCompress, 0.33.0"

open SharpCompress
open SharpCompress.Archives

let archive = ArchiveFactory.Open("test.7z")

for e in archive.Entries do
  if e.IsDirectory then
    printfn $"%A{e} is a directory"
  else
    printfn $"IsEncrypted %b{e.IsEncrypted}"
```

Perhaps the error message should be improved?

Can a directory be encrypted?

PR #768, opened 2023-09-23
Feature/761

Erior:
#761 

Triggered the report issue when traversing directories


adamhathcock:
Thanks.  I need to do a pass to remove `!` everywhere

PR #778, opened 2023-11-11
Throw ReaderCancelledException on reader cancelled

pathartl:
I needed a way to gracefully handle a cancellation. It would have been much easier to catch a specific type of exception instead of relying on the message of `InvalidOperationException`.

So, here you go!


adamhathcock (approved):

adamhathcock:
Thank you!

PR #780, opened 2023-11-14
Revert change disabling strong name signing in 92df1ec

### LINKED ISSUE #764: SharpCompress 0.34.0 does not have a strong name.
Hello,

First of all, thank you for your efforts at making this library.

As of the recent update that was released an hour long, SharpCompress 0.34.0 wasn't signed with a strong name, which causes every library that are signed with a strong key to emit warnings that say:

```
CS8002: Referenced assembly 'SharpCompress, Version=0.34.0.0, Culture=neutral, PublicKeyToken=null' does not have a strong name.
```

SharpCompress 0.33.0 didn't emit this warning because it was signed. Also, this signing key change wasn't listed in the changelogs.

What is the cause of the lack of signing? I would like this to be fixed.

Thanks!

caesay:
As title mentions. Strong name signing was disabled via 92df1ec in a large merge adding zstd support. 

Due to [this comment from @adamhathcock](https://github.com/adamhathcock/sharpcompress/issues/764#issuecomment-1723218890), I believe this was just a mistake. So here is a PR which resolves the issue. 

Accepting this PR will fix #764


adamhathcock:
Thank you!

Issue #783, opened 2023-12-01
NullReferenceException with RarReader since 0.33


IS4Code:
An error seems to have been introduced in 0.33 that affects reading some RAR archives using the `RarReader`.
Example RAR with the issue: https://telparia.com/fileFormatSamples/archive/rar/comment.rar
Even iterating through the archive throws an exception:
```cs
using var inputFile = File.Open("comment.rar", FileMode.Open);
var reader = RarReader.Open(inputFile);

while(reader.MoveToNextEntry()) ;
```
Exception:
```
   at SharpCompress.Common.Rar.RarVolume.<GetVolumeFileParts>d__11.MoveNext()
   at SharpCompress.Readers.Rar.RarReader.<GetEntries>d__13.MoveNext()
   at SharpCompress.Readers.AbstractReader`2.LoadStreamForReading(Stream stream)
   at SharpCompress.Readers.AbstractReader`2.MoveToNextEntry()
```
This works fine on 0.32.2, but does not work on 0.33 to 0.34.2.

IS4Code:
0.35 fixed this issue, but there is another null-related one with https://telparia.com/fileFormatSamples/archive/rar/password.rar on IReader.MoveToNextEntry:
```
System.ArgumentNullException: String reference not set to an instance of a String. (Parameter 's')

  Stack Trace: 
    Encoding.GetBytes(String s)
    UTF8EncodingSealed.GetBytes(String s)
    CryptKey5.GenerateRarPBKDF2Key(String password, Byte[] salt, Int32 iterations, Int32 keyLength)
    CryptKey5.Transformer(Byte[] salt)
    RarCryptoWrapper.ctor(Stream actualStream, Byte[] salt, ICryptKey key)
    RarHeaderFactory.TryReadNextHeader(Stream stream)
    RarHeaderFactory.ReadHeaders(Stream stream)+MoveNext()
    RarVolume.GetVolumeFileParts()+MoveNext()
    RarReader.GetEntries(Stream stream)+MoveNext()
    AbstractReader`2.LoadStreamForReading(Stream stream)
```

adamhathcock:
Looks like https://github.com/adamhathcock/sharpcompress/pull/798 will address

PR #784, opened 2023-12-04
Dont crash on reading rar5 comment #783

Erior:
Before we skipped the Service, this changed with #697 however, no tests was added and resulted in #783


adamhathcock (approved):

adamhathcock:
thanks!

Issue #782, opened 2023-11-21
Can not open tgz file


Broeckl:
Opening the attached file with this code fails:

                using (Stream stream = File.OpenRead(file))
                using (IReader reader = ReaderFactory.Open(stream))
                {
                    while (reader.MoveToNextEntry())
                    {
                        if (!reader.Entry.IsDirectory)
                        {
                            reader.WriteEntryToDirectory(outputFolder, new ExtractionOptions
                            {
                                ExtractFullPath = true, Overwrite = true
                            });
                        }
                    }
                }

An ArgumentException with "Value cannot be null. (Parameter 'path2')" occurs here:

   at System.IO.Path.Combine(String path1, String path2) in System.IO\Path.cs:line 316
   at SharpCompress.Common.ExtractionMethods.WriteEntryToDirectory(IEntry entry, String destinationDirectory, ExtractionOptions options, Action`2 write) in SharpCompress.Common\ExtractionMethods.cs:line 31
   at SharpCompress.Readers.IReaderExtensions.WriteEntryToDirectory(IReader reader, String destinationDirectory, ExtractionOptions options) in SharpCompress.Readers\IReaderExtensions.cs:line 32

The file can be opened with 7Zip. Your library can cope with comparable files. Any help is appreciated.

[03069284-040101.tgz](https://github.com/adamhathcock/sharpcompress/files/13423180/03069284-040101.tgz)

PR #785, opened 2023-12-04
Handle tar files generated with tar -H oldgnu that has large uid/gid values

Erior:
The tar file in #782 seems to be generated in oldgnu format, and sharpcompress failed to extract gid in the provided test, I have generated a general test with large value for both uid and gid using the normal setup of files.


adamhathcock:
thanks!

Issue #774, opened 2023-10-07
Failing to Read .7z Archive (LZMA:23)


Aragas:
I'm failing to extract this archive - "Sturgia is Norse for 1.0.2" via `ArchiveFactory.Open`
https://www.nexusmods.com/mountandblade2bannerlord/mods/2713?tab=files
Opening via 7z works, so the archive seems correct.
Not sure what data I can give to help - the archive is SOLID with 1 block and the method is LZMA:23
```
SharpCompress.Compressors.LZMA.DataErrorException: Data Error
   at SharpCompress.Compressors.LZMA.LzmaStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at SharpCompress.Compressors.LZMA.Utilites.Utils.ReadExact(Stream stream, Byte[] buffer, Int32 offset, Int32 length)
   at SharpCompress.Common.SevenZip.ArchiveReader.ReadAndDecodePackedStreams(Int64 baseOffset, IPasswordProvider pass)
   at SharpCompress.Common.SevenZip.ArchiveReader.ReadDatabase(IPasswordProvider pass)
   at SharpCompress.Archives.SevenZip.SevenZipArchive.LoadFactory(Stream stream)
   at SharpCompress.Archives.SevenZip.SevenZipArchive.LoadEntries(IEnumerable`1 volumes)
   at SharpCompress.Archives.AbstractArchive`2..ctor(ArchiveType type, SourceStream srcStream)
   at SharpCompress.Archives.SevenZip.SevenZipArchive..ctor(SourceStream srcStream)
   at SharpCompress.Archives.SevenZip.SevenZipArchive.Open(FileInfo fileInfo, ReaderOptions readerOptions)
   at SharpCompress.Factories.SevenZipFactory.Open(FileInfo fileInfo, ReaderOptions readerOptions)
   at SharpCompress.Archives.ArchiveFactory.Open(FileInfo fileInfo, ReaderOptions options)
   at SharpCompress.Archives.ArchiveFactory.Open(String filePath, ReaderOptions options)
   at ...
```

adamhathcock:
I guess this is an enhancement as this is a LZMA type not handled

Erior:
Have made some progress, this seems to be connected to LZMA EndOfStream marker, there are very small 7z files to test with that triggers the scenario. might take a look when I have the time.

pjominet:
I still seem to get a similar issue:
```
[...]

PR #786, opened 2023-12-09
LZMA EOS marker detection

Erior:
The DataError in #774 comes from the LZMA stream having EOS bytes at the end, we would need to check if the stream has it or not before raising DataError.

Added a test archive generated from 7za with the -meos flag set.


adamhathcock (approved):

adamhathcock:
thanks!

Issue #755, opened 2023-08-25
Rar5: IArchiveEntryExtensions WriteTo does not write to stream in 0.33 (RarArchive)


olegfeferman:
Hello,

I am trying to write a piece of code that verifies if user entered a correct password in case archive was password protected like this:

            using (var archive = RarArchive.Open(archiveFile, new ReaderOptions() { Password = pass }))
            {
                var entry = archive.Entries.FirstOrDefault(e => !e.IsDirectory);
                
                using (var stream = new MemoryStream( (int) entry.Size))
                {                    
                    entry.WriteTo(stream);                                        
                    if (stream.Length == 0)
                    {
                        ShowMessage("You have entered incorrect password.");
                    }
                }
            }

Problem is stream is always zero. Either it is a bug or I am doing something wrong.

olegfeferman:
It reports correct size in bytes of that entry

Erior:
I don't think you actually know if the password is correct until you try to decompress, that would give you a size, however it would be the wrong size / CRC, comparing the result with entry.Size would make more sense.
Btw, when I try this with encrypted header info, it throw an exception, you might need to catch that if you have such files.

olegfeferman:
Thank you for the feedback. When I test with Rar.encrypted_filesAndHeader.rar or Rar.encrypted_filesOnly.rar it works OK (in first case I get exception in second stream.Length is greater than 0). 
What I did not realize before you sent your message is that I am only having problem with Rar5.encrypted_filesAndHeader.rar and Rar5.encrypted_filesOnly.rar. Actually before I was just testing with a test archive that I created with latest version of Winrar (6.23).

adamhathcock:
RAR5 decryption is a known TODO: https://github.com/adamhathcock/sharpcompress#todos-always-lots

Dup of https://github.com/adamhathcock/sharpcompress/issues/372

olegfeferman:
I didn't know. Thank you!

Issue #718, opened 2023-01-09
SharpCompress.Common.InvalidFormatException: 'Unknown Rar Header: 43'


storm999:
I am having this error with rar5 format. My code works fine with zip and rar4.

adamhathcock:
There might be a new iteration in the RAR5 format.  Needs investigating and examples.

storm999:
I am unable directly attach a rar file to here.  Properties of the file shared in the picture. It is just a dummy rar of a simple txt file which was produced by latest edition of winrar.
![explorer_2023-01-10_21-38-04](https://user-images.githubusercontent.com/17099604/211634629-ca3766e0-5923-4a18-995f-7b49eb0b32d1.png)

Nanook:
Encryption is not supported

Issue #517, opened 2020-06-18
Unknown Rar Header: 117 when listing entries in a rar5 achive with encrypted file names.


evpxregu:
Steps to reproduce:
1) selected files to add to archive
2) select "set password"  in the archive creation screen
3) set password
4)make sure "encrypte file names" is checked)
5) in sharpcompress open the file with the correct password and try to access the Entries properties, an exception with unknown rar header: 117 will be thrown.

adamhathcock:
Decrypting RAR5 is a known TODO item on the README

PR #788, opened 2023-12-16
RAR5 decryption support

Erior:
Rar5 decryption support that handles our current test files, might work with issues

#755 #718 #517

Lets see if this passes CSharpier tests


adamhathcock:
Wow, thanks!  I know this is high on the list for lots of people

PR #794, opened 2024-01-02
Feature/rar5 blake2

Erior:
A Blake2 impl for RAR5, while testing it out I noticed that CRC and probably Blake2 for RAR5 with encryption is broken, the check is disabled for now when this is detectde. Exposes only the first 4 bytes for the Blake2 CRC externally to keep current API.


adamhathcock:
Better than nothing.  But yes, generally CRC is broken as I had no use for it when doing it orignally.

Thanks!

Issue #795, opened 2024-01-02
ReaderFactory not iterating over big entries


jbrockerville:
I've encountered an issue with the `ReaderFactory` and ZIP archives with "big" entries. If an archive has a sufficiently big enough entry, the reader fails to iterate over all the entries. No error. Just no next entry and drops out of the loop. The `ArchiveFactory` seems fine. Please see the attached code for generating the archive ("data/data.py") and reproducing the bug. Here is the output I get:
```
Entries via ReaderFactory...
  bin1.bin
Entries via ArchiveFactory...
  bin1.bin
  a.txt
  c.txt
```
[BigArchive.zip](https://github.com/adamhathcock/sharpcompress/files/13814299/BigArchive.zip)

adamhathcock:
Maybe  https://github.com/adamhathcock/sharpcompress/pull/799 will address

DannyBoyk:
Latest SharpCompress code produces identical results for ReaderFactory and ArchiveFactory, so it appears this can be closed out.

PR #799, opened 2024-01-04
Fix reporting size / position

Erior:
#795

We get out of sync after reporting the wrong size, mess up the 64 bit detection later


adamhathcock (approved):
Thanks!

Issue #771, opened 2023-09-26
Flush should probably not throw NotSupportedException


carlreinke:
According to the [Stream.Flush](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.flush) docs:
> In a class derived from [Stream](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream?view=net-7.0) that doesn't support writing, [Flush](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.flush?view=net-7.0) is typically implemented as an empty method to ensure full compatibility with other [Stream](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream?view=net-7.0) types since it's valid to flush a read-only stream.

https://github.com/adamhathcock/sharpcompress/blob/66c92637f9f5c07f9d644d40975833c6ba256190/src/SharpCompress/Compressors/LZMA/DecoderStream.cs#L17

adamhathcock:
Good catch and good PR

Kissaki:
I assume this ticket should have been closed as resolved by merged PR #801?

adamhathcock:
It has...not sure why it hasn't auto closed

ite-klass:
There was no closing keyword like resolves or fixes in the PR :)

PR #801, opened 2024-01-05
Issue 771, remove throw on flush for readonly streams

Erior:
#771 , remove throwing NotSupported for flush


adamhathcock:
Thanks!

PR #803, opened 2024-01-12
Add support for the UnixTimeExtraField in Zip files

### LINKED ISSUE #802: Add support for UnixTimeExtraField in Zips
We recently encountered a zip file that had the UnixTimeExtraField set in the local and central directory headers.  These times had the propery last modification time set for the files while the standard last modified time field in the headers was invalid.

This task is to add support for this extra field type.  Pull request incoming shortly.

DannyBoyk:
Fixes #802


adamhathcock (approved):
Thanks!

PR #805, opened 2024-01-26
Zip: Use last modified time from basic header when validating zip decryption

### LINKED ISSUE #804: ZIP: Use last modified time from basic header when validating zip decryption
When support was added for the UnixTimeExtraField a little while ago, zip decryption validation was broken if a zip is encrypted and has UnixTimeExtraFields present.  The zip encryption uses the basic header's last modified time value and not the times from the UnixTimeExtraField.

PR to fix incoming.

DannyBoyk:
The last modified time used for zip decryption validation must be the one from the basic header. If UnixTimeExtraFields are present, the previous implementation was attempting to verify against that value instead.
Fixed #804


adamhathcock:
Thanks!

PR #814, opened 2024-03-14
rar5 read FHEXTRA_REDIR and expose via RarEntry

coderb:
NOTE: api user should skip entries where RarEntry.IsRedir is true and not call OpenEntryStream()


adamhathcock:
This looks good, thanks!

This passed tests but are you saying the generic interface needs to change too?

coderb:
I haven't condidered that.  I am using RarReader.Open() which has proprerty `Entry` which is a `RarEntry`.

jgilbert2017:
hi- any chance this can be released to nuget?  thanks!

PR #816, opened 2024-03-14
rar5 improve memory usage

coderb:
use ArrayPool for stream buffer
use stackalloc for methods on file decompression code path


adamhathcock:
Thanks!

Stackalloc is troublesome for async usage but not sure async fits.

PR #817, opened 2024-03-17
#809 Add README.md to csproj for NuGet

btomblinson:
Linked the root README.md to the SharpCompress src project


adamhathcock (approved):
Thanks!

Issue #822, opened 2024-04-05
SharpCompress doesn't support ARM64 and RISCV filters for 7zip or XZ


klimatr26:
Hello.

7-zip added an ARM64 filter for ARM64 executables in version 23.01. It also added a RISCV filter in version 24.03.
![imagen](https://github.com/adamhathcock/sharpcompress/assets/154111881/74984522-e0a9-42a7-afe0-8ab494879ca1)

Currently, sharpcompress cannot decompress files that use those filters.

I was able to add support for those filters in 7-zip.

I'm not too familiar with XZ, but it uses liblzma, which incorporated those new filters. However, the code that sharpcompress uses for XZ filters seems to be based on the LZMA SDK.

Have a good day.

adamhathcock:
This was merged.  Thanks.

XZ is based on LZMA but I haven't refreshed myself to what is needed for XZ.  Will look more.

PR #823, opened 2024-04-07
Add support for 7z ARM64 and RISCV filters

klimatr26:
This allows sharpcompress to extract 7z archives that use ARM64 and RISCV filters, added in 7zip 23.01 and 24.03 respectively. This is related to #822


adamhathcock:
Thank you!

Issue #820, opened 2024-03-28
Tar corruption when writing active files


lopio:
I use the library to compress logs in a streaming manner, without saving a temporary file or storing to memory (RAM and diskspace restriction).

When I use the tar writer to archive active files (logs) the size of the tar header doesn't match the file in the tar, because the file has increased in size between the time the header was created and the file was fully read.

This result in the tar entry being corrupted, and sometime it can also result in the file in question not being recoverable.

Repro steps: Create a tar archive with a file that is being written in actively.

The fix is to stop reading the file at the size set by the tar header. Attached is the proposed fix as a git patch
[tarCorruption.patch](https://github.com/adamhathcock/sharpcompress/files/14793737/tarCorruption.patch)

adamhathcock:
I seemingly can't use that patch file 

Any chance of a PR?

lopio:
I can't find a way to create a branch for the PR. I do have my complete git diff here:
```
PS C:\git\sharpcompress> git diff
diff --git a/src/SharpCompress/Utility.cs b/src/SharpCompress/Utility.cs
index 9471c3e..0493e84 100644
--- a/src/SharpCompress/Utility.cs
+++ b/src/SharpCompress/Utility.cs
@@ -270,16 +270,27 @@ public static class Utility
         return sTime.AddSeconds(unixtime);
     }

-    public static long TransferTo(this Stream source, Stream destination)
+    public static long TransferTo(this Stream source, Stream destination, Int64? size = null)
     {
+        Int64 realSize = size ?? source.Length;
         var array = GetTransferByteArray();
         try
         {
             long total = 0;
             while (ReadTransferBlock(source, array, out var count))
             {
-                total += count;
-                destination.Write(array, 0, count);
+                if (total + count > realSize)
+                {
+                    int bytesToWrite = (int) (realSize - total);
+                    destination.Write(array, 0, (int) bytesToWrite);
+                    total = realSize;
+                    break;
+                }
+                else
+                {
+                    total += count;
+                    destination.Write(array, 0, count);
+                }
             }
             return total;
         }
diff --git a/src/SharpCompress/Writers/Tar/TarWriter.cs b/src/SharpCompress/Writers/Tar/TarWriter.cs
index 2427db3..53ce0ea 100644
--- a/src/SharpCompress/Writers/Tar/TarWriter.cs
+++ b/src/SharpCompress/Writers/Tar/TarWriter.cs
[...]

PR #825, opened 2024-04-09
Fix tar corruption when sizes mismatch

adamhathcock:
Addresses https://github.com/adamhathcock/sharpcompress/issues/820


src/SharpCompress/Utility.cs
```diff
@@ -269,16 +269,25 @@ public static DateTime UnixTimeToDateTime(long unixtime)
...
         {
             long total = 0;
             while (ReadTransferBlock(source, array, out var count))
             {
-                total += count;
-                destination.Write(array, 0, count);
+                if (size is not null && total + count > size)
+                {
+                    var bytesToWrite = (int)(size - total);
+                    destination.Write(array, 0, bytesToWrite);
```
lopio: This will throw a ArgumentOutOfRangeException if total > size
I do not see an end condition it which we stop reading on we read the desired size. I originally used a break as a end condition, but  see line 278 for alternative fix

src/SharpCompress/Utility.cs
```diff
@@ -269,16 +269,25 @@ public static DateTime UnixTimeToDateTime(long unixtime)
...
     }
 
-    public static long TransferTo(this Stream source, Stream destination)
+    public static long TransferTo(this Stream source, Stream destination, long? size = null)
     {
         var array = GetTransferByteArray();
         try
         {
             long total = 0;
             while (ReadTransferBlock(source, array, out var count))
```
lopio, line 278: There should be an additional exit condition for the loop

while (ReadTransferBlock(source, array, out var count) && (size is null || total < size))

lopio (changes requested):
There is no end condition to stop the reding until there is an exception

src/SharpCompress/Utility.cs
```diff
@@ -277,8 +277,33 @@ public static long TransferTo(this Stream source, Stream destination)
...
+        try
+        {
+            long total = 0;
+            var remaining = size;
+            while (ReadTransferBlock(source, array, remaining, out var count))
+            {
                 destination.Write(array, 0, count);
+                total += count;
+                remaining -= count;
+                if (remaining - count < 0)
```
lopio, line 307: This should be before the remaining -= count;
For exemple, if we have a remaining 15 bytes, and a count of 10. It will put remaining to 5 after the write. But 5 < 10 so it will never read the last 5 bytes.

lopio (changes requested):

src/SharpCompress/Utility.cs
adamhathcock: done

src/SharpCompress/Writers/Tar/TarWriter.cs
```diff
@@ -91,7 +91,13 @@ public void Write(string filename, Stream source, DateTime? modificationTime, lo
         header.Size = realSize;
         header.Write(OutputStream);
 
-        size = source.TransferTo(OutputStream);
+        if (realSize >= int.MaxValue)
```
lopio: why restrict the size? Wouldn't be better to keep the length to a long like in it used too. I can see a use case for a 2GB file being tar the 7zipped

adamhathcock: the `Write` command only accepts up to an int.  I could subtract and manage it but that's a future PR if it's a problem.

lopio: I don't see why this is an issue. The write only takes limited amount (81920 bytes) using the transfer array created in ReadTransferBlock. The file will be separated into parts for reading then writing to tar separately.

I compressed in a tar files bigger then 2 GB using the tar writer before

adamhathcock: you're right...overall size has nothing to do with transfer size/array

brain fart

src/SharpCompress/Utility.cs
```diff
@@ -277,8 +277,38 @@ public static long TransferTo(this Stream source, Stream destination)
...
+            }
+            while (ReadTransferBlock(source, array, maxReadSize, out var count))
+            {
                 destination.Write(array, 0, count);
+                total += count;
+                if (remaining - count < 0)
+                {
+                    break;
+                }
+                remaining -= count;
```
lopio, line 311: missing maxReadSize adjustment. This reintroduces the bug that used to be there
if (remaining < maxReadSize)
{
    maxReadSize = (int)remaining;
}

src/SharpCompress/Utility.cs
```diff
@@ -277,8 +277,38 @@ public static long TransferTo(this Stream source, Stream destination)
...
+
+    public static long TransferTo(this Stream source, Stream destination, long maxLength)
+    {
+        var array = GetTransferByteArray();
+        try
+        {
+            long total = 0;
+            var remaining = maxLength;
+            var maxReadSize = array.Length;
+            if (remaining < maxLength)
```
lopio: This conditions is impossible because remaining = maxLength
It should be if (remaining < maxReadSize )

lopio (changes requested):

src/SharpCompress/Utility.cs
adamhathcock: you're right

src/SharpCompress/Utility.cs
adamhathcock: you're right

lopio (approved):

PR #835, opened 2024-04-24
Prevent infinite loop when reading corrupted archive

### LINKED ISSUE #736: TarHeaderFactory Infinite loop

Blokyk:
Fixing another infinite loop when reading corrupted tar archives. Tries to fix #736


adamhathcock:
thanks!

PR #852, opened 2024-06-22
Fixed extractions after first ZIP64 entry is read from stream

pathartl:
When a ZIP64 entry is extracted from an archive, the header for the next entry is read out of order. This commit reads the header again, then resets the position of the RewindableStream back to the beginning of the entry, just after the 4 byte header signature. This ensures the correct header signature data and reader (in the correct position) are fed to ReadHeader. Previously this would have read the data at an incorrect offset and would not yield a viable ZipHeader.

I haven't done extensive testing with this nor did I write any automated tests.


pathartl:
Not sure why tests are failing here... is `format` formatting issues? Running all tests through VS pass.

adamhathcock:
format is using csharpier to enforce a code style.  The dotnet tool is available locally and `dotnet csharpier .` should fix the formatting

adamhathcock:
Thanks!

PR #854, opened 2024-07-06
return Stream.Null when 7z entry has no stream

zgabi:
bug: 7z file extraction fails when the archive contains empty file

I'm using the following code to extract a 7z file:
```
using var sevenZip = SharpCompress.Archives.SevenZip.SevenZipArchive.Open(data);
sevenZip.WriteToDirectory(root);

```

But it fails when the archive file (`data`) contains an empy file.
The same `GetCompressedStream` method returns `Stream.Null`, I think this method should also return `Stream.Null` instead throwing an exception.

It would be good to put an empy file to the tests\TestArchives\Original folder and to all the test files, to cover this case in the tests, but I don't know how to create them.


adamhathcock:
There should be a tested added but can merge just for consistency

zgabi:
Yes. There should be an empty file in the tests\TestArchives\Original folder... but I don't know how to add that to all "exotic" archives in the tests\TestArchives\Archives folder (for example tests\TestArchives\Archives\7Zip.ZSTD.7z). So please add it. Thank you.

Issue #837, opened 2024-04-26
False positive in tar detection


DaleStan:
When I call `ArchiveFactory.IsArchive` with the attached file (which started life as a SQL CE database), it incorrectly reports that it's a tar file.

Is it possible to make the detection code stricter, or have I just stumbled across a pathological case?
[false positive.tar.txt](https://github.com/adamhathcock/sharpcompress/files/15133633/false.positive.tar.txt)

adamhathcock:
You might have stumbled on something because the file gets read and looks for a tar header.

PR #855, opened 2024-07-10
Check crc on tar header

Erior:
#837

We could use the CRC for checking validation of the header, however it would seem the Tar.Empty.tar test case is invalid


tests/SharpCompress.Test/Tar/TarArchiveTests.cs
```diff
@@ -217,8 +217,7 @@ public void Tar_Empty_Archive()
     {
         var archiveFullPath = Path.Combine(TEST_ARCHIVES_PATH, "Tar.Empty.tar");
         using Stream stream = File.OpenRead(archiveFullPath);
-        using var archive = ArchiveFactory.Open(stream);
-        Assert.True(archive.Type == ArchiveType.Tar);
+        Assert.Throws<InvalidOperationException>(() => ArchiveFactory.Open(stream));
```
adamhathcock: Feels like not being able to open an empty tar is a regression

adamhathcock (changes requested):

tests/SharpCompress.Test/Tar/TarArchiveTests.cs
Erior: You are correct, the empty file is a special case that needs to be taken care of, updated solution proposal

Erior:
The last could actually be change to return crc == chk1 || crc == chk2   ... instead of the if... if you prefere that I can change

adamhathcock (approved):

adamhathcock:
Thanks!

PR #859, opened 2024-07-25
Fix #858 - Replaces invalid filename characters

### LINKED ISSUE #858: unicode characters are causing file to Currupt at extraction
https://www.deviantart.com/zenoasis/art/Japanese-TV-Dorama-folder-icon-pack-162-1077192465
Download zip from there 
image

Extract using SharpCompress,
we'll see that the files `REAL 恋愛殺人捜査班 : Real -  Renai Satsujin Sosa Han.png` and `あの子の子ども : My Girlfriend's Child.png` are extracted as 0 byte file with following names `REAL µüïµä¢µ«║Σ║║µì£µƒ╗τÅ¡ ` , `πüéπü«σ¡Éπü«σ¡Éπü¿πéÖπéé`

I also tried with IBM437 encoding but same result.

However when you extract using 7zip you can see that it extracts fine and 7zip makes some changes to file name - which seeems to be removing `:` character which might be  either `U+A789` or `U+2236`

filenames from 7zip extraction `REAL 恋愛殺人捜査班 _ Real -  Renai Satsujin Sosa Han.png`, `あの子の子ども _ My Girlfriend's Child.png`

DineshSolanki:
Add a method to replace invalid characters in file names with underscores during file extraction. This prevents errors related to invalid file names.

### DIFF
--- a/src/SharpCompress/Common/ExtractionMethods.cs
+++ b/src/SharpCompress/Common/ExtractionMethods.cs
@@ -37,6 +37,7 @@ internal static class ExtractionMethods
         options ??= new ExtractionOptions() { Overwrite = true };
 
         var file = Path.GetFileName(entry.Key.NotNull("Entry Key is null")).NotNull("File is null");
+        file = Utility.ReplaceInvalidFileNameChars(file);
         if (options.ExtractFullPath)
         {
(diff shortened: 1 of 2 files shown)


adamhathcock:
Thanks!

PR #862, opened 2024-08-06
Embed debugging symbols

lostmsu:
This greatly simplifies debugging of SharpCompress source because it removes the need to setup nuget symbol downloading.

It also simplifies debugging scenarios for anyone who is not using NuGet for some reason (e.g. git submodule, or another build system).


adamhathcock:
This is the a good idea.

Any chance this can include changes to get the build on nuget have three green checks? https://nuget.info/packages/SharpCompress/0.37.2

I was going for https://learn.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg with snupkg but they're out of favor?

lostmsu:
TBH I don't want to figure out the build system.

For regular .NET project I'd just add `/p:ContinuousIntegrationBuild=true` to the Github actions file. But to do it cleanly with what I saw one would need to create a build option, that the target would convert to this string, and then in GitHub actions set the option.

adamhathcock:
I think https://github.com/adamhathcock/sharpcompress/pull/864 does more.  I still don't do releases with GA so I just dumped `ContinuousIntegrationBuild` in it

adamhathcock:
Thanks for this.  Unfortunately I have to merge manually so it's done in a separate PR

PR #864, opened 2024-08-08
Update csproj to get green marks and update deps

adamhathcock:
Should replace https://github.com/adamhathcock/sharpcompress/pull/862


lostmsu:
I am for changing symbols type to embedded though, so IMHO take both.

PR #876, opened 2024-10-23
Restore stream position in ArchiveFactory.IsArchive

### LINKED ISSUE #870: ArchiveFactory.IsArchive does not rewind the input stream's position
Calling `ArchiveFactory.IsArchive` with a stream as input parameter will potentially seek the given stream. That is necessary of course, but after returning, the stream should be seeked back to its original position.

After all, you might want to check whether a given input is a valid archive that can be handled by SharpCompress before continuing to construct an archive from that stream (or do other things with that stream).

It looks like `ArchiveFactory.FindFactory` already has that logic but is only exposed via `ArchiveFactory.Open`, which in itself seeks the input stream by opening the archive immediately (maybe that should also be changed? not sure).

Morilli:
- closes #870

By resetting the stream position after the archive check, the stream position will be restored to its original value after the method returns.

### DIFF
--- a/src/SharpCompress/Archives/ArchiveFactory.cs
+++ b/src/SharpCompress/Archives/ArchiveFactory.cs
@@ -189,9 +189,10 @@ public static class ArchiveFactory
 
         foreach (var factory in Factory.Factories)
         {
+            var isArchive = factory.IsArchive(stream);
             stream.Position = startPosition;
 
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):

adamhathcock:
Thanks!

PR #877, opened 2024-10-26
Fixed bug in zip time header flags

StarkDirewolf:
Parsing the times in the zip headers worked, but the flag check was wrong, so only modified time was ever parsed. Corrected this in a more readable way.

Also stopped exceptions being thrown if there was an issue during this header parsing, returning null values instead.

Remarks have been added to highlight that they return UTC `DateTime`'s, rather than `LastModifiedTime` which returns a local `DateTime`.


adamhathcock:
Looks good, thanks!

StarkDirewolf:
Ah, thanks for letting me know :) Pushed the styling change.

PR #878, opened 2024-10-29
Fix XZBlock padding calculation when its stream's starting position % 4 != 0

### LINKED ISSUE #873: Padding calculation wrong when starting position % 4 != 0
```
private void SkipPadding()
    {
        var bytes = (int)(BaseStream.Position % 4);
```

I have a BaseStream where the xz file doesn't start at position 0. If this position is not dividable by 4, then the padding calculation is wrong.  
Please fix this by calculating the position from the starting position.
Thanks

Morilli:
- closes #873

I've also removed the unused `_bytesRead` field that seems to have last been touched in #300.


Morilli:
Hm, test failures seem to occur because `XZBlock` is instantiated with a (RewindableStream) stream that throws an exception when accessing its `Position`... not sure how valid that even is, considering reading the block would eventually fail anyways because `SkipPadding` already accesses the stream's position because this change.

adamhathcock (approved):

adamhathcock:
Thanks!

PR #884, opened 2024-12-18
Exports unclutter

YoshiRulz:
Moves some types from the `SharpCompress` namespace to `SharpCompress.Helpers`, so they won't be used accidentally by consumers. (Builds upon #579.) And removes the redundant `ReadOnlyCollection<T>`.


adamhathcock:
Looks great, thanks!

PR #892, opened 2025-01-28
don't run net48 on non-windows

adamhathcock:


YoshiRulz:
If this is because `dotnet test` can't find Mono, a workaround is to downgrade the Ubuntu container image: https://github.com/TASEmulators/BizHawk/blob/61ce55ab4c67b788fe9fe4e01db2e242441bbbb3/.github/workflows/ci.yml#L35

PR #894, opened 2025-02-11
Fix condition in rar v3 code

### LINKED ISSUE #890: Unknown Rar Header: 238
Exception:
```
SharpCompress.Common.InvalidFormatException: Unknown Rar Header: 238
   at SharpCompress.Common.Rar.Headers.RarHeaderFactory.TryReadNextHeader(Stream stream)
   at SharpCompress.Common.Rar.Headers.RarHeaderFactory.ReadHeaders(Stream stream)+MoveNext()
   at SharpCompress.Common.Rar.RarVolume.GetVolumeFileParts()+MoveNext()
   at SharpCompress.Readers.Rar.RarReader.GetEntries(Stream stream)+MoveNext()
   at SharpCompress.Readers.AbstractReader`2.NextEntryForCurrentStream()
   at SharpCompress.Readers.AbstractReader`2.MoveToNextEntry()
```

Code:
```
using var fs = File.Open("F:\\Rar.rar", FileMode.Open);
using var reader = ReaderFactory.Open(fs);
{
}
```

Happens when moving to the 8th track in [archive](https://vk.com/doc322150273_437251882?hash=zE9Yxg87k384IbiReOcv0vvAiShZw6J4ngpE83PY0jP&dl=WueLvzsvyFCInvr1zuowseaFrt8bvUTlgZ1u96wyreT&api=1&no_preview=1)

No issues with winrar

Morilli:
- closes #890

This was an interesting one to track down. Basically, when `UnpWriteData` is called with a size of `0`, it must call `writeStream.Write` even if `destUnpSize` is 0 to ensure that `Suspended` is set to false.
The same condition change must be done here: https://github.com/adamhathcock/sharpcompress/blob/94789ce455cf88cbe0636abc0071c1c03d4a2ebf/src/SharpCompress/Compressors/Rar/UnpackV1/Unpack.cs#L266 
to make sure that the code passes over both `if (destUnpSize < 0)` and `if (suspended)` conditions and continues below.

While investigating this bug I found some other issues in the rarVM code, but I'm not sure how to test them and whether that code is even still used nowadays.

TODO: Find some example archive that shows this bug that isn't 40MB and add a test case for it.

### DIFF
--- a/src/SharpCompress/Compressors/Rar/UnpackV1/Unpack.cs
+++ b/src/SharpCompress/Compressors/Rar/UnpackV1/Unpack.cs
@@ -713,6 +713,8 @@ internal sealed partial class Unpack : BitInput, IRarUnpack, IDisposable
 
     private void UnpWriteData(byte[] data, int offset, int size)
     {
+        // allow destUnpSize == 0 here to ensure that 0 size writes
+        // go through RarStream's Write so that Suspended is set correctly
         if (destUnpSize < 0)
         {
             return;
--- a/src/SharpCompress/Compressors/Rar/UnpackV1/Unpack.cs
+++ b/src/SharpCompress/Compressors/Rar/UnpackV1/Unpack.cs
@@ -263,7 +263,7 @@ internal sealed partial class Unpack : BitInput, IRarUnpack, IDisposable
             if (((wrPtr - unpPtr) & PackDef.MAXWINMASK) < 260 && wrPtr != unpPtr)
             {
                 UnpWriteBuf();
-                if (destUnpSize <= 0)
+                if (destUnpSize < 0)
                 {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


adamhathcock (approved):
Seems valid to me, not sure what to do to ensure this doesn't happen again.  Tests or comments?

Morilli:
A test would be great but I don't know how to generate an archive that would exhibit this problem, I only used the archive provided in #890 to test this.

I can add a comment to `UnpWriteData`, however looking at it, the "new" 2017 rar5 code seems to have this same issue, so maybe it needs to be changed  there as well?

adamhathcock:
I'm imagining it has the same issue but I'm not an expert with it

adamhathcock:
either this PR (or possible the OpenRead) PR has seemingly broke a disposing test on the reader?

Morilli:
welp, it looks like applying this same diff on the rar5 code breaks it... not very reassuring.

adamhathcock (approved):

PR #896, opened 2025-02-11
Rar2 v20,v26 Multimedia (Audio) decoder fix

Nanook:
This is a fix for the Audio Decoder in unpack20 (v20/v26). When decoding a rar that uses the old multimedia compression algorithm an error is raised:

```
Object reference not set to an instance of an object.
   at SharpCompress.Compressors.Rar.UnpackV1.UnpackUtility.makeDecodeTables(Span`1 lenTab, Int32 offset, Decode dec, Int32 size) in /_/SharpCompress/Compressors/Rar/UnpackV1/UnpackUtility.cs:line 197

```

This is because the MultiDecode[] objects are never instantiated.

`private readonly MultDecode[] MD = new MultDecode[4];`  (Unpack20.cs:17)

I looked back through the repo hoping to find and old working commit, but they have never been set. I even dug out the original Unrar project (pre SharpCompress) and that was the same.

This fixed the crash, but the decompression still wasn't correct. After many hours, I decided to compare the code with the C++ implementation used by 7zip and finally tracked it down to 2 bad casts in the Audio Decoder - (byte) when it should actually be (sbyte). Fortunately the rest of the code seemed to work perfectly after that (big relief).

I've not got any small archives that display this behaviour for use with unit tests. Current tests pass and this fix is being tested on many archives.

:-)

### DIFF
--- a/src/SharpCompress/Compressors/Rar/UnpackV1/Unpack20.cs
+++ b/src/SharpCompress/Compressors/Rar/UnpackV1/Unpack20.cs
@@ -14,7 +14,13 @@ namespace SharpCompress.Compressors.Rar.UnpackV1;
 
 internal partial class Unpack
 {
-    private readonly MultDecode[] MD = new MultDecode[4];
+    private readonly MultDecode[] MD = new[]
+    {
+        new MultDecode(),
+        new MultDecode(),
+        new MultDecode(),
+        new MultDecode()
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):

adamhathcock:
thanks!

Issue #399, opened 2018-07-08
Really bad performance.


SuperJMN:
I'm using this code. 

It takes like 10 minutes to extract 272 files from a 7-zip archive that is not special by any means. Well, it's been compressed using the Ultra setting… :)

I has some of them that are about 80MB in size.

What could be wrong?

Thanks!!

```
public class DriverPackageInstaller : IDriverPackageInstaller
    {
        public async Task InstallDriverPackage(string fileName, string destination)
        {
            using (var package = SevenZipArchive.Open(fileName))
            {
                await Extract(package.Entries, "Cityman", @"NewPack\Pre-OOBE");
                await Extract(package.Entries, "PostOOBE", @"NewPack\Post-OOBE");
            }
        }

        public static async Task Extract(ICollection<SevenZipArchiveEntry> entries, string sourceFolder, string destinationFolder)
        {
            var translatedPath = sourceFolder.Replace("\\", "/");

            foreach (var entry in entries.Where(x => x.Key.StartsWith(translatedPath) && !x.IsDirectory))
            {
                await ExtractEntryToFile(entry, sourceFolder, destinationFolder);
            }
        }

        private static async Task ExtractEntryToFile(IArchiveEntry entry, string sourceFolder, string destination)
        {
            using (var input = entry.OpenEntryStream())
            {
                var path = entry.Key.Replace($"{sourceFolder}/", "");
                var fileName = Path.GetFileName(path);
                var directoryName = Path.GetDirectoryName(path);
                var newFolder = Path.Combine(destination, directoryName);
                var outputPath = Path.Combine(newFolder, fileName);

                Directory.CreateDirectory(newFolder);
                using (var output = File.OpenWrite(outputPath))
                {
                    await input.CopyToAsync(output);
                }                
            }
        }
    }
```

adamhathcock:
The 7zip implementation is far from optimized.  I’m sure some quick wins could be from pooling byte arrays but I haven’t looked into it.

SuperJMN:
Thank you. I hope it gets better. I'm using it inside this application: https://github.com/SuperJMN/Lumia-WoA-Installer (in the most recent dev branch, still not committed). I will use it to import driver packages. They're .7z files with Ultra compressing level.

SuperJMN:
It's already merged. I hope it's published to NuGet.org soon! Thank you.

adamhathcock:
https://www.nuget.org/packages/sharpcompress/0.22.0

[...]

PR #898, opened 2025-02-16
Implement ReadByte for LzmaStream and LzOutWindow

Morilli:
Similar to #897, this improves performance and reduces memory allocations when reading certain 7-zip files.

Using the .7z file from https://github.com/adamhathcock/sharpcompress/issues/399#issuecomment-405028725, I measured a speedup in reading / skipping the entire .7z file from ~23 seconds to ~18 seconds, and memory allocations dropped by ~8.5GB.


adamhathcock (approved):

adamhathcock:
awesome, thanks!

PR #900, opened 2025-02-28
make WriteToDirectory functions use ExtractAllEntries

### LINKED ISSUE #710: Need to improve the decompression performance of 7-Zip
image
image
This file is taking an unbelievable amount of time to decompress, and it's getting slower and slower,
I can't even get it to finish decompressing,
Is there something wrong here?

Morilli:
- closes #710 and possibly some other related ones

For extracting files from solid archives calling `ExtractAllEntries` is required to get acceptable extraction performance and should therefore always be called.

Clearly people are not aware of this function, and the base library not using it is not helping. Perhaps it should be mentioned in the `USAGE.md`?

### DIFF
--- a/src/SharpCompress/Archives/ArchiveFactory.cs
+++ b/src/SharpCompress/Archives/ArchiveFactory.cs
@@ -122,10 +122,7 @@ public static class ArchiveFactory
     )
     {
         using var archive = Open(sourceArchive);
-        foreach (var entry in archive.Entries)
-        {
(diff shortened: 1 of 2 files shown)


adamhathcock:
Thanks.  It probably does make sense for this change.

I've had my head out of this game for a while so I'm not surprised somethings are done inefficiently.  

Notes about USAGE should be updated as well.

PR #901, opened 2025-03-07
Handle XZ CheckType SHA-256

ms264556:
Add `CheckType.SHA256` to the `XZHeader.BlockCheckSize` 'calculation' so XZ can decompress e.g. RPM package payloads.


adamhathcock (approved):
Thanks!

PR #903, opened 2025-03-09
Base Reader implementation of .ARC format

TwanVanDongen:
Implemented handling of .ARC format, initially only 'uncompressed' streams.


TwanVanDongen:
still more algorithms to implement though ;)

PR #904, opened 2025-03-09
Provide access to extended attributes for 7-zip

jdpurcell:
Some 7-zip archives have extended attributes to store Unix permissions. This caused problems in the past so the extended attributes [were stripped](https://github.com/adamhathcock/sharpcompress/pull/153) when reading. However, this information can be useful or even critical in some cases, such as when extracting Linux/macOS prebuilt binaries for developing with Qt. These archives have execute permissions set for some files, and some of the files are intended to be created as symbolic links.

This PR provides access to the original full attributes via a new property in `SevenZipEntry`. The other changes were just small cleanups. Specifically, as I looked at `AttribDefined`/`SetAttrib` and wondered if there should be extended versions of these, I saw that these are properties in an internal class an nothing is ever referencing them. I also noticed `CrcDefined` had only one reference, and around where it was used showed an IDE warning that went away by using `is not null` instead, so I thought it would be better to remove it.

image

With this new `ExtendedAttrib` property, I was able to write an [extraction routine](https://github.com/jdpurcell/naqt/blob/7cdba606333b6cfabec00676151aa31a6e35cdf2/Helpers/General.cs#L52) that allows the Qt 7-zip archives to be extracted in a fully working state.

### DIFF
--- a/src/SharpCompress/Common/SevenZip/ArchiveReader.cs
+++ b/src/SharpCompress/Common/SevenZip/ArchiveReader.cs
@@ -996,6 +996,11 @@ internal class ArchiveReader
                             numFiles,
                             delegate(int i, uint? attr)
                             {
+                                // Keep the original attribute value because it could potentially get
+                                // modified in the logic that follows. Some callers of the library may
+                                // find the original value useful.
+                                db._files[i].ExtendedAttrib = attr;
+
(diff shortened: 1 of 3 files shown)


adamhathcock:
Thank you!

PR #905, opened 2025-03-11
ARC decompression methods 3 and 4 added

TwanVanDongen:


adamhathcock (approved):
Thanks!

src/SharpCompress/IO/ReadOnlySubStream.cs
```diff
@@ -31,7 +31,7 @@ public ReadOnlySubStream(Stream stream, long? origin, long bytesToRead)
 
     public override void Flush() { }
 
-    public override long Length => throw new NotSupportedException();
```
Morilli, line 34: I don't think this is correct. The length of a substream is not equal to the length of the underlying `base` stream.

adamhathcock: actually, you're right.....I need a test or something to ensure things

PR #907, opened 2025-03-19
Optimize LZ OutWindow.CopyBlock

jdpurcell:
After a bit of profiling, this function seemed like a good candidate for optimization. Testing on a [large-ish archive](https://download.qt.io/online/qtsdkrepository/mac_x64/desktop/qt6_682/qt6_682/qt.qt6.682.clang_64/6.8.2-0-202501260836qtdeclarative-MacOS-MacOS_14-Clang-MacOS-MacOS_14-X86_64-ARM64.7z) (412 MB), I'm seeing extraction times like:
|CPU|Before|After|% of Original|
|-|-|-|-|
|Apple M3|31.1 s|26.8 s|86%|
|Core i7-6700k|37.2 s|33.8 s|91%|

### DIFF
--- a/src/SharpCompress/Compressors/LZMA/LZ/LzOutWindow.cs
+++ b/src/SharpCompress/Compressors/LZMA/LZ/LzOutWindow.cs
@@ -85,28 +85,56 @@ internal class OutWindow
         _streamPos = _pos;
     }
 
-    public void CopyBlock(int distance, int len)
+    public void CopyPending()
     {
-        var size = len;
-        var pos = _pos - distance - 1;
-        if (pos < 0)
+        if (_pendingLen < 1)
         {
-            pos += _windowSize;
+            return;
         }
-        for (; size > 0 && _pos < _windowSize && _total < _limit; size--)
+        var rem = _pendingLen;
+        var pos = (_pendingDist < _pos ? _pos : _pos + _windowSize) - _pendingDist - 1;
+        while (rem > 0 && HasSpace)
         {
             if (pos >= _windowSize)
             {
                 pos = 0;
             }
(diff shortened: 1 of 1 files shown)


jdpurcell:
After more testing I found that my initial commit despite helping a lot with the Qt archive, could slow down others. An example is [here](https://github.com/staxrip/staxrip/releases/download/v2.44.5/StaxRip-v2.44.5-x64.7z) (426 MB). I investigated the differences and determined that `CopyBlock` is called in two different ways, one directly, and one indirectly via `CopyPending`. Despite my best attempts, I couldn't figure out a way to avoid splitting them up. It's very counterintuitive, it almost seems like `CopyPending` could benefit more from this optimization since it's often copying more data on average, but I think the issue is about 7% of the time in my testing it gets called with a 1-length request, whereas `CopyBlock` never does. And there's no way to work around the performance hit with `if`s; the only thing I can guess is that 7% is enough to really mess up the branch prediction. And for whatever reason, archives like the Qt one are biased way more to calling `CopyBlock` vs `CopyPending`, whereas some others are quite the opposite.

As much as I hated having to duplicate some of the code, I did my best to clean it up at the same time so it's not too bad overall in my opinion. After splitting it up like this, I still get the big performance gains from the Qt archive, I found a couple others that were quicker too, and then another handful including the one I just linked that at least don't regress in performance anymore (if anything maybe a slight increase but it's probably a drop in the bucket). Despite the perhaps-not-as-simple code, I feel it's worth it for the performance potential. If you agree, feel free to merge, and if not, no worries - I am already using a custom build and could continue to do so.

adamhathcock (approved):
Copying for a reason is always good.

adamhathcock:
Thanks!

PR #909, opened 2025-03-22
Update USAGE.md to remove problematic extraction example

Morilli:
I've modified the "extract all files" example to use `IReader.ExtractAllEntries`. As per the added comment this must be done for solid archives and probably has better performance compared to iterating over entries in non-sequential order for other archives too.

I've added a new example for iterating over entries that doesn't extract them.


adamhathcock (approved):
Thanks for doing this!

PR #912, opened 2025-03-25
Optimize BufferedSubStream.ReadByte

jdpurcell:
In `BufferedSubStream`, by changing what `BytesLeftToRead` and `_cacheLength` track, we can simplify the work to be done when reading a byte. `BytesLeftToRead` now tracks the amount of completely unread/uncached data, and `_cacheLength` tracks the total cache length including the consumed portion. Then  we only need to update the cache offset when reading a byte. It also moves the logic to refresh the cache into its own method. Interestingly, only about half of the performance increase comes from simplifying the variable usage. The other half comes from moving out the cache refresh logic, presumably because it makes the `ReadByte` method small enough for callers to inline.

Core i7-6700k (1.7% reduction):
| Method | Mean    | Error    | StdDev   |
|------- |--------:|---------:|---------:|
| Before  | 1.088 s | 0.0031 s | 0.0028 s |
| After  | 1.070 s | 0.0011 s | 0.0010 s |

Apple M3 (2.8% reduction):
| Method | Mean     | Error   | StdDev  |
|------- |---------:|--------:|--------:|
| Before  | 785.0 ms | 1.60 ms | 1.42 ms |
| After  | 763.2 ms | 2.84 ms | 2.52 ms |

Time to extract a 14MB 7z. The benchmarks are with .NET 8.0 but it also helped slightly with .NET Framework 4.8 (~1.1% reduction).

I wish I could keep going with these optimization PRs but I think I'm out of ideas after this one 😄.

### DIFF
--- a/src/SharpCompress/IO/BufferedSubStream.cs
+++ b/src/SharpCompress/IO/BufferedSubStream.cs
@@ -20,7 +20,7 @@ internal class BufferedSubStream(Stream stream, long origin, long bytesToRead)
 
     public override void Flush() { }
 
-    public override long Length => BytesLeftToRead;
+    public override long Length => BytesLeftToRead + _cacheLength - _cacheOffset;
 
     public override long Position
     {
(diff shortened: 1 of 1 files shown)


src/SharpCompress/IO/BufferedSubStream.cs
```diff
@@ -20,29 +20,41 @@ internal class BufferedSubStream(Stream stream, long origin, long bytesToRead)
 
     public override void Flush() { }
 
-    public override long Length => BytesLeftToRead;
+    public override long Length => BytesLeftToRead + _cacheLength - _cacheOffset;
```
jdpurcell, line 23: It seems weird that `Length` in a stream refers only to the remaining portion, but since that's the way it already worked I assume it's for a reason.

src/SharpCompress/IO/BufferedSubStream.cs
```diff
@@ -20,29 +20,41 @@ internal class BufferedSubStream(Stream stream, long origin, long bytesToRead)
...
 
     public override long Position
     {
         get => throw new NotSupportedException();
         set => throw new NotSupportedException();
     }
 
+    private void RefillCache()
+    {
+        var count = (int)Math.Min(BytesLeftToRead, _cache.Length);
```
jdpurcell, line 33: This is a new bit of logic here; the old code would always request `_cache.Length` bytes when refilling. It probably wasn't hurting anything but I figure it's better to request only what we need.

src/SharpCompress/IO/BufferedSubStream.cs
adamhathcock: maybe that's not the contract but I "meant" to do it 😬

adamhathcock (approved):
Smells good, thanks!

PR #913, opened 2025-03-26
Fix regression with BufferedSubStream calculation

jdpurcell:
Fixes regression introduced in #912. Most of my test archives didn't trigger this code path so I didn't notice until now.


adamhathcock (approved):

adamhathcock:
thanks!

PR #916, opened 2025-05-14
Implement multipart rar handling for ExtractAllEntries

### LINKED ISSUE #915: ExtractAllEntries doesn't follow the same logic as Entries for detecting multi part rar files
Hello there,

I'm trying to use `ExtractAllEntries` due to `Entries` having performance issue by the suggestion found [here](https://stackoverflow.com/a/44379540).

However, `ExtractAllEntries` won't work with multi part rar files.

Here is a simple example to demo the issue:

```csharp
using System;
using System.Linq;
using SharpCompress.Archives;

namespace SharpCompressIssue;

class Program {
    static void Main(string[] args) {
        var archive = ArchiveFactory.Open("a.part01.rar");
        Console.WriteLine("Entries:");
        foreach (var entry in archive.Entries.Where(entry => !entry.IsDirectory)
                     .Select(entry => entry.Key)) {
            Console.WriteLine($"\t{entry}");
        }

        Console.WriteLine("ExtractAllEntries:");
        using var reader = archive.ExtractAllEntries();

        while (reader.MoveToNextEntry()) {
            var entry = reader.Entry;
            if (!entry.IsDirectory) {
                Console.WriteLine($"\t{entry.Key}");
            }
        }
    }
}
```

It will give the result like this:

```
Entries:
        Complete.png
        Magic.png
        Summon.png
        Support.png
ExtractAllEntries:
Unhandled exception. SharpCompress.Common.MultiVolumeExtractionException: Streamed archive is a Multi-volume archive.  Use different RarReader method to extract.
   at SharpCompress.Readers.Rar.SingleVolumeRarReader.ValidateArchive(RarVolume archive)
   at SharpCompress.Readers.Rar.RarReader.GetEntries(Stream stream)+MoveNext()
   at SharpCompress.Readers.AbstractReader`2.LoadStreamForReading(Stream stream)
   at SharpCompress.Readers.AbstractReader`2.MoveToNextEntry()
   at SharpCompressIssue.Program.Main(String[] args) in /Users/jingbian/Projects/KifaNet/Experimental/SharpCompressIssue/Program.cs:line 19
```

You can find the full package [here](https://drive.google.com/file/d/1gXYUoouy7tX-tEKEI-nrce-yc5DE1vEr/view?usp=sharing) or you can also create the multi part rar files.

Morilli:
- closes #915.

One side-effect of the changes here is that *skipping over encrypted entries for which the password is not known in a multipart rar file using the reader interface* no longer works.

However *reading* those entries does not work because it isn't currently implemented, and the skipping was bad as well because it would fail to skip all file parts and returned after skipping the first part only, so I think it is fair to remove this functionality.
It could be implemented properly but I think it would take some effort to extract the needed information and perhaps need additional interface methods.

### DIFF
--- a/src/SharpCompress/Archives/Rar/RarArchive.cs
+++ b/src/SharpCompress/Archives/Rar/RarArchive.cs
@@ -67,6 +67,16 @@ public class RarArchive : AbstractArchive<RarArchiveEntry, RarVolume>
 
     protected override IReader CreateReaderForSolidExtraction()
     {
+        if (this.IsMultipartVolume())
+        {
+            var streams = Volumes.Select(volume =>
+            {
+                volume.Stream.Position = 0;
+                return volume.Stream;
+            });
+            return RarReader.Open(streams, ReaderOptions);
+        }
+
         var stream = Volumes.First().Stream;
         stream.Position = 0;
         return RarReader.Open(stream, ReaderOptions);
--- a/src/SharpCompress/Readers/AbstractReader.cs
+++ b/src/SharpCompress/Readers/AbstractReader.cs
@@ -131,7 +131,7 @@ public abstract class AbstractReader<TEntry, TVolume> : IReader, IReaderExtracti
     {
         var part = Entry.Parts.First();
 
-        if (!Entry.IsSolid && Entry.CompressedSize > 0)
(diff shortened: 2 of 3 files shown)


adamhathcock:
Thanks.  I think it's far that you can't skip passworded entries.  SOLID archives are supposed to be a continuous stream which needs to be decrypted.

PR #921, opened 2025-07-06
Fix volume FileName property potentially missing

### LINKED ISSUE #920: Volume's FileName doesn't work when readerOptions is specified
I'm trying to get FileName of archive parts with

```cs
        var archive = ArchiveFactory.Open("a.part01.rar", new ReaderOptions {
            Password = null,
            ArchiveEncoding = new ArchiveEncoding(Encoding.UTF8, Encoding.UTF8)
        });
        foreach (var volume in archive.Volumes) {
            Console.WriteLine($"Volume {volume.Index}: {volume.FileName}");
        }
```
It gives me empty file names while the Index can be retrieved normally.

If I remove the options like

```cs
        var archive = ArchiveFactory.Open("a.part01.rar");
        foreach (var volume in archive.Volumes) {
            Console.WriteLine($"Volume {volume.Index}: {volume.FileName}");
        }
```
It works properly.

Expectation: It should return the relevant archive file names in both cases.

Example project can be found here: https://drive.google.com/file/d/1knv9PH24TkHipKwXrL7hmxth96thxXxb/view?usp=sharing

Morilli:
- closes #920

If the `ReaderOptions` has `LeaveStreamOpen` set (which is set to true by default), the stream used to report a volume's `FileName` was a `NonDisposingStream` instead of the originally passed in stream, which never worked.

### DIFF
--- a/src/SharpCompress/Common/Volume.cs
+++ b/src/SharpCompress/Common/Volume.cs
@@ -7,12 +7,14 @@ namespace SharpCompress.Common;
 
 public abstract class Volume : IVolume
 {
+    private readonly Stream _baseStream;
     private readonly Stream _actualStream;
 
     internal Volume(Stream stream, ReaderOptions? readerOptions, int index = 0)
     {
         Index = index;
         ReaderOptions = readerOptions ?? new ReaderOptions();
+        _baseStream = stream;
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):
Thanks!

PR #924, opened 2025-07-13
Fix 7-zip solid archive detection

### LINKED ISSUE #923: Archive-Level SevenZip "IsSolid" flag always set to false if a 7Z archive only has one solid block.
SharpCompress's archive-level SevenZip "IsSolid" flag is always set to false if a 7Z archive only has one solid block, while 7z (the program) is able to still identify the archive as solid. Not the hugest concern, but I wanted to make an issue for documentation purposes.

Morilli:
- closes #923 and fixes `IsSolid` logic in general

For some reason 7-zip doesn't like to create 1-block solid archives so I had to create a new 7z file with one less file included than the `7Zip.solid.7z` to get a solid archive with only one block.


adamhathcock (approved):
Thank you!

HeroponRikiBestest:
Thank you!

PR #928, opened 2025-07-19
fix DotSettings options to conform to current code style and editorconfig

Morilli:
Some of these generate very annoying code, especially because the editorconfig disagrees with it.


adamhathcock (approved):
Thanks

Issue #927, opened 2025-07-14
SharpCompress fails to validate tar compressed using zstd


SimonCahill:
It's been a while since I've actively used this library, but I'd had good experiences in the past, so I decided to use it once more.

I have the following test case:

```cs
[TestMethod]
public async Task TestCompressDirectory_ZstCompressedTar_FileInRoot() {
    var tmpFile = Path.GetTempFileName();
    var fileBytes = GenerateRandomMemory(4096);
    File.WriteAllBytes(tmpFile, fileBytes.ToArray());

    var tmpPath = new DirectoryInfo(Path.Combine(Path.GetTempPath(), "HermodTests", "EmailArchiverTests", nameof(TestCompressDirectory_ZstCompressedTar_FileInRoot)));
    tmpPath.Create();

    File.Move(tmpFile, Path.Combine(tmpPath.FullName, Path.GetFileName(tmpFile)));

    await m_emailArchiver.CompressDirectoryAsync(tmpPath, CompressionType.Squeezed);
    var expectedPath = Path.Combine(tmpPath.Parent?.FullName ?? string.Empty, $"{nameof(TestCompressDirectory_ZstCompressedTar_FileInRoot)}.tzst");

    Assert.IsTrue(File.Exists(expectedPath), $"Archive file ({expectedPath}) was not created.");

    using var fStream = File.OpenRead(expectedPath);

    if (fStream.Length < 512) {
        // This is likely not a valid Zstandard compressed file.
        Assert.Fail($"Archive file ({expectedPath}) is not a valid Zstandard compressed file.");
    }

    Assert.IsTrue(VerifyTarHeader(), $"Archive file ({expectedPath}) is not a valid TAR file after decompression.");

    fStream.Seek(0, SeekOrigin.Begin);
    using var zstDecompStream = new DecompressionStream(fStream);
    using var tarArchive = TarArchive.Open(zstDecompStream);

    var entries = tarArchive.Entries.ToList(); // HERE!

    Assert.IsTrue(entries.Any(), $"Archive file ({expectedPath}) is empty.");

    bool VerifyTarHeader() {
        using var zstDecompStream = new DecompressionStream(fStream);

        Assert.IsTrue(zstDecompStream.CanRead, $"Archive file ({expectedPath}) is not a valid Zstandard compressed stream.");
        return TarArchive.IsTarFile(zstDecompStream);
    }
}
```

Interestingly, it fails to detect the tar header in the marked line.

[...]

PR #931, opened 2025-07-21
Added ArgumentException to Archive.Open implementations

SimonCahill:
This minor change affects the `Open` methods of each Archive type.

The change is only a condition, determining whether or not the `Stream` is seekable, and throws an `ArgumentException` if that's the case.

I'd though of implementing this, too, but @Morilli gave me the final motivation.


adamhathcock (approved):
Thank you!

SimonCahill:
No worries.

This also addresses #927.

PR #935, opened 2025-08-01
Rewind buffer fix for directory extract.

Nanook:
Just a little bug fix. New tested added that fails prior to fix.


adamhathcock (approved):
Thanks!

PR #943, opened 2025-09-23
ZStandard tar support

### LINKED ISSUE #941: zstd support for tar (can potentially make a PR)
I noticed zstd is not supported for tar but this library does include the zstd c# library for a select few other functions.

I was able to work around this fairly easily by just creating the zstd stream myself to pass to the reader factory, but I am not sure if there is a reason not to do something similar in the official code.  Given the ability to rewind the sharpcompress stream wrapper and the fact IReader is meant as forward only it would seem like this should work but I don't know the internals of sharpcompress well enough to say if I am missing something.  If not happy to do a PR to add zstd as a tar supported format.

Here is my current code:

```csharp
var entries = new SortedSet<ArchiveEntry>();
Stream stream = File.OpenRead(archivePath);
toDispose.Add(stream);
if (archivePath.EndsWith("tar.zstd", StringComparison.CurrentCultureIgnoreCase) || archivePath.EndsWith("tar.zst") || archivePath.EndsWith(".tzst")|| archivePath.EndsWith(".tzstd")) {
				stream = new ZstdSharp.DecompressionStream(stream);
				toDispose.Add(stream);
}

if (reader == null)
				reader = ReaderFactory.Open(stream);
```

mitchcapper:
Closes #941 looked at implementing ireader but would have to use some of the much more low level Zstandard classes to read the frames directly and it didn't look conducive to that.


adamhathcock:
Thanks for this.  Please format with csharpier

mitchcapper:
done

adamhathcock:
Thank you!

PR #945, opened 2025-09-24
Extension hinting for ReaderFactory for better first try factory success

### LINKED ISSUE #942: optional smarter readerfactory testing through readeroptions adding an extension or ArchiveType for hinting?
Right now factories are just bruteforced in order when detecting an unknown stream.   While things can have the wrong extension or the user may not know the extension I would assume a good bit of time the user does have the filename.

The user could optionally provide this for hinting.  We already maintain a list of supported extensions for external use ( IFactory.GetSupportedExtensions ) that we could then make use of internally.   This would be largely backwards compatible as we would simply switch which factory we try first to be the suspected factory, if that fails the normal bruting of the other factories would be used.   Things like TarFactory could use the ReaderOptions extension to even decide what compression algorithm to try first (although that would require adding an internal table of extension to compression alg as I don't think we maintain that data anywhere).

If this sounds sane I can do a PR.

mitchcapper:
Also used by TarFactory to hint what compressiontype to attempt first

Close #942 

This adds extension hinting support to ReaderFactory and allows compressors to optionally use the hinting as well for their own prioritization.

This only prioritizes which factories/methods we try first so should not result in a change of functionality for anyone (it also only does the prioritization if you provide the extension as a hint).

I added prioritization to tar as well, this required rewriting it a bit but it allowed removing a good bit of duplicate code.  I don't know why certain tar methods require wrapping the rewind stream again, would seem this maybe should be a property on the type, but I did preserve this functionality. 

I added gzip to TarFactory as well.  While the GzipFactory would eventually handle it from brute force it seemed like it belonged in TarFactory with the others.  

For bare tar archives I no longer use TarReader.Open as it has redundant code once again brute forcing the compression types.  Given the code path was only triggered if the IsTarFile check passed before even if they were slightly different it shouldn't matter.

I don't think this introduces much of a different code path that could be exploited but the tar change would probably be the biggest.

This makes it a bit easier to track 'legitimate' exceptions in the library when debugging vs ones caused during the type brute forcing.

### DIFF
--- a/src/SharpCompress/Factories/TarFactory.cs
+++ b/src/SharpCompress/Factories/TarFactory.cs
@@ -1,10 +1,13 @@
+using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using SharpCompress.Archives;
 using SharpCompress.Archives.Tar;
 using SharpCompress.Common;
 using SharpCompress.Compressors;
 using SharpCompress.Compressors.BZip2;
+using SharpCompress.Compressors.Deflate;
 using SharpCompress.Compressors.LZMA;
 using SharpCompress.Compressors.Lzw;
 using SharpCompress.Compressors.Xz;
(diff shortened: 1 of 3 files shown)


mitchcapper:
No more so trying to debug sharp compress some of the exceptions types thrown were too generic to just ignore for the entire library but would always occur at startup when it bruted some of the formats.

PR #1017, opened 2025-11-14
Fix some IStreamStack and SharpCompressStream functions

### LINKED ISSUE #960: ExtractAllEntries returns no entries when reading a ZIP from a Stream
Appears to be a regression in 0.41.0

[test.zip](https://github.com/user-attachments/files/22993107/test.zip)

Generated via ```7z a test.zip wallhaven-225935.jpg```. Example input file must be sufficiently large for the bug to become apparent.

The following code produces 0 results in SharpCompress 0.41.0.

```csharp
using var zipStream = File.OpenRead(@"C:\test.zip"); // Important! don't use ArchiveFactory.Open(string) overload
using var archive = SharpCompress.Archives.ArchiveFactory.Open(zipStream);
using var reader = archive.ExtractAllEntries();
{
    Console.WriteLine(reader.Entry);
}
```

I bisected and the issue starts occurring here https://github.com/adamhathcock/sharpcompress/commit/76de7d54c7eacf014d3fafc8352f61f3e32ef11d (replacing `stream.Position = 0;` with `((IStreamStack)stream).StackSeek(0);`)

Morilli:
- resolves #960

Buffer handling seems to have been completely broken.
I've fixed the handling in `IStreamStack.StackSeek` that basically duplicated the logic from `SharpCompressStream.Seek` (while not actually working correctly), fixed incorrect behavior of the `IStreamStack.BufferPosition` setter and fixed and off-by-one in `SharpCompressStream.Seek`.

I suspect there are more issues in `StackSeek` that may or may not even be relevant for how buffered `IStreamStack`s are even used. It's hard to tell because the method documentation does not seem to fully match the implementation.

This allows reverting #964 now that the underlying issue has been fixed.

TODO:
- [x] make some tests for this to validate behavior

### DIFF
--- a/src/SharpCompress/IO/IStreamStack.cs
+++ b/src/SharpCompress/IO/IStreamStack.cs
@@ -167,7 +167,8 @@ namespace SharpCompress.IO
             if (lastBufferingIndex != -1)
             {
                 var bufferingStream = stack[lastBufferingIndex];
-                var targetBufferPosition = position - bufferingStream.GetPosition() + bufferingStream.BufferPosition;
+                var targetBufferPosition =
+                    position - bufferingStream.GetPosition() + bufferingStream.BufferPosition;
 
                 if (targetBufferPosition >= 0 && targetBufferPosition <= bufferingStream.BufferSize)
                 {


Morilli:
> Seek and setting BufferPosition should be the same thing?

For a buffered stream, it is the same if the seek would only seek to somewhere inside of the currently buffered region. Unfortunately some buffered streams aren't seekable so we can't unconditionally call seek without trying to set the buffer position beforehand. This seems to be mainly required for archive detection logic as that uses a buffered, non-seekable stream for rewind logic where we need to be able to rewind the buffer position while not having access to the `Seek` method.

PR #1019, opened 2025-11-14
ARJ's methods 1, 2 and 3 implemented for streaming

TwanVanDongen:
Addition of CompressedMost, Compressed and CompressFastest algorithms.


TwanVanDongen:
bug found, need to fix first

TwanVanDongen:
It can be integrated now, so please do; but know that some more enhancements are still underway, which can be either incorporated now or in future. Would be nice if we can add current functionality.

PR #1028, opened 2025-11-20
Buffer boundary tests

TwanVanDongen:
Checking whether the algorithms handle input stream chunking of 128kb correctly


adamhathcock (approved):

adamhathcock:
Thanks!

Issue #1001, opened 2025-10-30
Support multi-threading when supplying a path/FileInfo


adamhathcock:
Since the ultimate source is a local file, there's no reason multi-threading won't work for some formats.  What is required:

- a format that supports header reading separate from bytes or one that can skip entries (e.g. Zip, Rar, Tar)
- compressed entries can be read from separate parts simultaneously (probably most?)
- known non-supporting as things are single compressed streams that usually wrap a Tar (Gzip, BZip2, etc)
- unknown: "SOLID" support.  SOLID Rars and 7ZIps compress headers with entry bytes.  7Zip might be able to work but differently.  If the entire Rar is SOLID then it's like a Gzipped Tar which can't really work either.

adamhathcock:
Need an example of how this can work per stream/file.  This is still single threaded decompression but can do multiple entries at once.

What compressions support multi-threaded usage?

PR #1040, opened 2025-11-27
Add multi-threading support for File based single volume Zips and Rars

adamhathcock:
Starts https://github.com/adamhathcock/sharpcompress/issues/1001

After thinking a while, SourceStream gives all the info I need.

related: https://github.com/adamhathcock/sharpcompress/issues/1042


adamhathcock:
@copilot why do some tests fail?

adamhathcock:
@copilot find out why the tests fail and only on windows

Issue #203, opened 2016-12-20
Detect if rar file is password protected?


evpxregu:
Hi,

Is there any way to check if a rar or another other type is password protected? 
Currently I check if it throws a CryptographicException but this does not work for .rar files where the filenames are not encrypted, when I then try to open the entrystream or extract the entries I get a NullReferenceException.

Thanks!

adamhathcock:
I might be doing something wrong for that case.  I think the only encryption flag I check for RARs is on the ArchiveHeader.

I'd have to dig into the spec.

Could be a good PR.

Nanook:
I noticed this issue and found that checking IsEncrypted on the Entry was the best method. If you're not using passwords then you shouldn't attempt to process an Entry with IsEncrypted is True.

PR #1060, opened 2025-12-02
Add archive-level password protection flags for 7z and rar

HeroponRikiBestest:
Adds an encrypted flag for 7z archives and rar archives. The main reason/utility of this PR is because ExtractAllEntries is (practically) needed for solid 7z and rar archives, but an exception will be thrown if any of the files are password protected. This avoids the current requirement to pre-iterate through file entries to check for password protection.

This implementation mainly copies the logic for 7z's IsEncrypted flag. It checks the first file in the archive and checks whether it's encrypted. While it's technically possible that some later files aren't encrypted; as mentioned previously, the main utility of this flag is so that the aforementioned additional logic can be avoided for solid archives.

This also lines up with https://github.com/adamhathcock/sharpcompress/issues/203#issuecomment-1117164160 , where it was suggested that the most reliable way to check for password protection is to check the entry anyways.

### DIFF
--- a/src/SharpCompress/Archives/Rar/RarArchive.cs
+++ b/src/SharpCompress/Archives/Rar/RarArchive.cs
@@ -84,10 +84,7 @@ public class RarArchive : AbstractArchive<RarArchiveEntry, RarVolume>
 
     public override bool IsSolid => Volumes.First().IsSolidArchive;
 
-    public override bool IsEncrypted =>
-        Entries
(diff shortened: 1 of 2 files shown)


adamhathcock:
looks like you need to run `csharpier`

HeroponRikiBestest:
Ran csharpier on all changed files, let me know if I missed anything

PR #1063, opened 2025-12-06
Add alternative option for writing TAR archives with USTAR header format

drone1400:
**Changes Summary**
- Adds an option when creating a TarWriter to write the TAR headers using the USTAR format
- Change is implemented by adding an enum property to TarWriterOptions
- Does not change default behavior, unless manually specified when creating a TarWriter object via the TarWriterOptions, will use the existing modern GNU TAR header

**Issues**
- Perhaps, that I did not add any unit test for validating the USTAR headers. They should be fine since I have been using these changes for almost a year with no issue...

**Context**
Some time last year when I was using SharpCompress for a game mod manager, I ran into an issue where tar.bz2 files compressed with SharpCompressed were incompatible with some source ports of that game that would only accept the old USTAR format.

I didn't think of making a pull request at the time as it seems like a very niche thing, but now I'm thinking someone else might find it useful too.

### DIFF
--- a/src/SharpCompress/Common/Tar/Headers/TarHeader.cs
+++ b/src/SharpCompress/Common/Tar/Headers/TarHeader.cs
@@ -10,7 +10,10 @@ internal sealed class TarHeader
 {
     internal static readonly DateTime EPOCH = new(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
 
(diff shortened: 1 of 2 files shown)


adamhathcock:
This is great, thanks!

If you could format it as I'll add a test or two with copilot.

adamhathcock (approved):

adamhathcock:
Thanks!

PR #1096, opened 2026-01-03
RAR async

adamhathcock:
This should make tests for asynchronous overloads use the `AsyncOnlyStream` for RAR and force all RAR paths to use asynchronous methods to read from streams


adamhathcock:
@copilot ensure the PR achieves it's goal by making the tests pass

adamhathcock:
@copilot make the async rar tests pass.  Don't fix other tests.  LoadStreamForReadingAsync isn't loading the enumerator so make it load correctly

Issue #1115, opened 2026-01-07
the zip file with '\\' in it's entry.Key make error when run in linux


jackletter:
Although the zip specification stipulates that file path names should use '/' as the separator, there are still non-compliant writes that can cause issues when extracting on Linux. Could we consider uniformly replacing '\' with '/' after decoding?

Additional note: I tried implementing the replacement in the CustomDecoder, but if the zip file specifies an encoding, SharpCompress won't invoke the CustomDecoder, preventing me from making the corrections.

```csharp
        if (Flags.HasFlag(HeaderFlags.Efs))
        {
            Name = ArchiveEncoding.DecodeUTF8(name);
            Comment = ArchiveEncoding.DecodeUTF8(comment);
        }
        else
        {
            Name = ArchiveEncoding.Decode(name);
            Comment = ArchiveEncoding.Decode(comment);
        }
```

adamhathcock:
It something making a zip with `\` in the path in SharpCompress?  If something is doing this, I don't want to account for it with code.

Maybe the encoding method should have more options?

jackletter:
Users may use various programs to compress ZIP file packages, which is beyond our control. As a service provider, we must strive to be as compatible as possible with users' files. @adamhathcock

jackletter:
The simplest approach is to make the ArchiveEncoding.DecodeUTF8 method virtual. This way, I can inherit from ArchiveEncoding and override DecodeUTF8 to internally replace '\\' with '/'.

adamhathcock:
> Users may use various programs to compress ZIP file packages, which is beyond our control. As a service provider, we must strive to be as compatible as possible with users' files. [@adamhathcock](https://github.com/adamhathcock)

Disagree.  Keys are dumb strings and whatever produces should confirm to spec.  If the programs don't, it's not up to SharpCompress to fix it.  People can do things themselves with SC though.

> The simplest approach is to make the ArchiveEncoding.DecodeUTF8 method virtual. This way, I can inherit from ArchiveEncoding and override DecodeUTF8 to internally replace '\' with '/'.

It is the simplest, however might not be the best.

jackletter:
So please merge the pull request, so I won't have to duplicate SharpCompress source code internally.

PR #1117, opened 2026-01-07
Change ArchiveEncoding to interface.

adamhathcock:
Simplify class.  Question what to do about Forced and complex access.  Was custom encoder used?

Should I have the interface use a `Func` then default to what I use but allow for anything?

Started by https://github.com/adamhathcock/sharpcompress/issues/1115


adamhathcock:
This is instead of https://github.com/adamhathcock/sharpcompress/pull/1116

Issue #1105, opened 2026-01-04
7Zip extraction performs significantly worse in 0.43.0


Camble:
As the title suggests, 7Zip extraction 0.43.0 is significantly slower than 0.41.0.

I extract the contents in memory for later writing to disk

<img width="936" height="92" alt="Image" src="https://github.com/user-attachments/assets/7811a89f-5795-4314-97b8-e29e8bec849d" />

**Benchmark sample:**
```csharp
FileStream fileStream = File.OpenRead(filename);
IArchive archive = ArchiveFactory.Open(fileStream);
IReader reader = archive.ExtractAllEntries();
while (reader.MoveToNextEntry())
{
  // etc...
  EntryStream source = reader.OpenEntryStream();
  // etc...
}
```

rube200:
I also noticed it from 0.42.0 to 0.43.0
On my case it is extremely slow. 
(I have a lot of small files)

adamhathcock:
I'm looking to narrow this down but I think I found a disposal issue that caused a lot of allocations because LZMA never returned some memory to a pool.

adamhathcock:
Try https://www.nuget.org/packages/SharpCompress/0.44.0-beta.33

If that's better, I'll push it out has a hotfix to the release

Camble:
Allocations are much lower, but the duration is unchanged.

<img width="641" height="66" alt="Image" src="https://github.com/user-attachments/assets/ce4d382a-4b85-4fd4-9a13-6863c2219eed" />

adamhathcock:
I'll have to have a closer look and think as to what changed to make it slower.  Having done perf testing and look at the slowest bits, they haven't changed for a while I think.

julianxhokaxhiu:
I did try to look a bit into this so far but unfortunately without any meaningful impact.

The curious thing is that I did manage to generate a repro archive where I can "feel" the slowness, however if I run the unit test used to read `7zip.solid.7z` the difference is very minimal ( ~2s ).

Here the snippet to generate the test archive ( you need powershell and 7z installed in the system )

**512MB** ( LZMA2:24 - 16MB dictionary - 1 block )
```pwsh
[...]

PR #1163, opened 2026-01-27
[WIP] Improve I/O operations while reading big solid archives

julianxhokaxhiu:
This PR is a tentative effort to improve the extraction time for 7z archives, especially ones that are compressed as 1 block solid using 16MB dictionary.

The main idea here is to reduce the number of access to cache ( while doing Skip ), improve performance in writing files to the disk ( use 1MB buffer by default ) and aggressively inline codepaths that are reused across executions ( avoiding jumps in the code but instead favoring a more linear execution at the expense of code size ).

Feel free to test it and report back. In my own case I can finally see the extraction of files in the archive mentioned in #1105

Below Claude Sonnet 4.5 summary

---

### Problem
7Zip extraction with large solid archives (16MB dictionaries, 1-block compression) was extremely slow in version 0.42.0+ compared to 0.41.0, taking hours instead of minutes on high-end hardware.

### Root Causes Identified Through Profiling
1. **StreamExtensions.Skip()** consuming 94.54% CPU - byte-by-byte reading via CopyTo(Stream.Null)
2. **Excessive Path.GetFullPath() calls** - Called 3+ times per extracted file
3. **Small I/O buffers** - 80KB buffers insufficient for modern NVMe drives
4. **Method call overhead** - Hot-path LZMA decompression methods not inlined

#### 1. Skip Operation Optimization
**File**: StreamExtensions.cs
- Changed from `ReadOnlySubStream + CopyTo(Stream.Null)` (byte-by-byte) to buffered reading
- Increased buffer size from implicit default to **1MB** using ArrayPool
- Added fast path for `BufferedSubStream` to skip via internal method
- **Impact**: Reduced Skip CPU usage from 94.54% → 82.24% → negligible

**File**: BufferedSubStream.cs
- Added `SkipInternal()` method with 1MB buffer for efficient skipping
- Skips cached data instantly, uses large buffered reads for remainder
- **Impact**: Eliminates repeated RefillCache() calls when skipping large amounts of data

#### 2. Increased I/O Buffer Sizes (80KB → 1MB)
**Files Modified**:
- IArchiveEntryExtensions.cs - BufferSize constant
- AbstractReader.cs - CopyTo/CopyToAsync calls

**Rationale**: Modern NVMe Gen 5 drives benefit significantly from larger sequential I/O operations
- **Impact**: Better disk I/O throughput, reduced system calls

#### 3. FileStream Optimization
**File**: IArchiveEntryExtensions.cs
- Replaced `File.Open()` with explicit `FileStream` constructor
- Specified 1MB buffer size explicitly
- Added `useAsync: true` for async operations to enable overlapped I/O on Windows
- **Impact**: Better async I/O performance, reduced context switching

#### 4. Path Processing Optimization
**File**: ExtractionMethods.cs
- **Cached `entry.Key`** to avoid multiple property accesses
- **Reduced Path.GetFullPath() calls** from 3 per file to 1 by consolidating path operations
- **Combined Path.Combine calls**: `Combine(folder, file)` instead of separate operations
- **Moved security validation** before filesystem calls to avoid unnecessary I/O
- **Impact**: WriteEntryToDirectory CPU reduced from 84.95% → 31.74% (63% reduction)

#### 5. LZMA Decompression Micro-optimizations
**Files Modified**:
- RangeCoder.cs
- RangeCoderBit.cs  
- LzmaDecoder.cs

**Added `[MethodImpl(MethodImplOptions.AggressiveInlining)]` to hot-path methods**:
- `RangeCoder.Decoder`: GetThreshold, Decode, DecodeBit, DecodeDirectBits
- `BitDecoder.Decode` (called millions of times)
- `LzmaDecoder.LenDecoder.Decode`
- `LzmaDecoder.Decoder2`: DecodeNormal, DecodeWithMatchByte
- `LzmaDecoder.LiteralDecoder`: GetState, DecodeNormal, DecodeWithMatchByte

**Rationale**: These methods are in tight decompression loops; inlining reduces call overhead and enables cross-method JIT optimizations
- **Impact**: Reduced LZMA decompression overhead, though fundamental decompression work remains CPU-intensive as expected

### Performance Results
- **Skip operation**: No longer a bottleneck (was 94.54% CPU)
- **Path processing**: 63% reduction in WriteEntryToDirectory overhead
- **Overall extraction**: Significantly faster, especially noticeable with large solid archives
- **Hot path**: Now dominated by actual LZMA decompression work (unavoidable)

### Technical Notes
- All buffer allocations use `ArrayPool<byte>` to minimize GC pressure
- Changes maintain backward compatibility
- Security validations (path traversal checks) preserved
- Code formatted with CSharpier per project standards

### Testing
Tested with 700MB+ solid 7zip archive with 16MB dictionary on AMD 9800X3D with NVMe Gen 5 drive. Extraction time improved from hours to minutes, matching performance of version 0.41.0.

### DIFF
--- a/src/SharpCompress/Archives/IArchiveEntryExtensions.cs
+++ b/src/SharpCompress/Archives/IArchiveEntryExtensions.cs
@@ -9,7 +9,7 @@ namespace SharpCompress.Archives;
 
 public static class IArchiveEntryExtensions
 {
-    private const int BufferSize = 81920;
+    private const int BufferSize = 1048576; // 1MB buffer for better disk I/O performance
 
     /// <param name="archiveEntry">The archive entry to extract.</param>
     extension(IArchiveEntry archiveEntry)
@@ -135,7 +135,14 @@ public static class IArchiveEntryExtensions
                 options,
                 (x, fm) =>
                 {
-                    using var fs = File.Open(destinationFileName, fm);
+                    // Use larger buffer for better disk I/O performance
+                    using var fs = new FileStream(
+                        destinationFileName,
+                        fm,
+                        FileAccess.Write,
+                        FileShare.None,
+                        bufferSize: 1048576
+                    ); // 1MB buffer
                     entry.WriteTo(fs);
                 }
             );
@@ -155,7 +162,15 @@ public static class IArchiveEntryExtensions
                     options,
                     async (x, fm, ct) =>
                     {
-                        using var fs = File.Open(destinationFileName, fm);
+                        // Use async I/O with large buffer for better performance
+                        using var fs = new FileStream(
+                            destinationFileName,
+                            fm,
+                            FileAccess.Write,
+                            FileShare.None,
+                            bufferSize: 1048576,
+                            useAsync: true
+                        ); // 1MB buffer
                         await entry.WriteToAsync(fs, null, ct).ConfigureAwait(false);
                     },
                     cancellationToken
--- a/src/SharpCompress/Common/ExtractionMethods.cs
+++ b/src/SharpCompress/Common/ExtractionMethods.cs
@@ -27,7 +27,6 @@ internal static class ExtractionMethods
         Action<string, ExtractionOptions?> write
     )
     {
-        string destinationFileName;
         var fullDestinationDirectoryPath = Path.GetFullPath(destinationDirectory);
 
         //check for trailing slash.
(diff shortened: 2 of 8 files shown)


julianxhokaxhiu:
I'll take care of Copilot suggestions tomorrow. Thanks

adamhathcock:
There's probably a few things here that might need to be separate PRs to be consumable.

1) making buffer size an option (ideally pooling) which I want to do
2) improvements for copying/Skip extension methods
3) improvements to LZMA in general

I definitely want to do them, but just break them down.  It helps me as I'm in the middle of a large refactor to get async working all the way.

I'll make a stab at #1 today

adamhathcock:
I just realized this PR is against `master` which is currently dev.  You probably want `release` so we can get these kind of fixes out faster

julianxhokaxhiu:
I'll have to rework the entire patch then as the PR states there are 249 file changes now which is not what I want. I didn't knew `release` is the branch currently used to release current nuget artifacts. I'll update this to point again to master and refine the patch, and eventually see how we can backport that to release after if it's ok with you. Let me know, cheers

adamhathcock:
I think the release branch is now have buffer size centralized into a single static which should make things easier.

I'll back port this to master which should have further changes later

https://github.com/adamhathcock/sharpcompress/pull/1165

julianxhokaxhiu:
I'll close this PR then in favor of yours, so we keep this one only as a ref.

PR #1180, opened 2026-02-02
Mark as IsAotCompatible

mcraiha:
For .NET 8 and .NET 10

I tested with `dotnet build` and `dotnet pack` from **src** folder. Also I ran `dotnet test` without issues. Tested with Ubuntu 24.04 + .NET 10 under WSL.


adamhathcock:
Added to release, hope it works

adamhathcock:
in master too

PR #1227, opened 2026-02-21
Added Net7.0 / Net 6.0 / Net 5.0 and NetStandard2.1

Nanook:
Minimal change, thought I might as well. Great progress recently. Nice one.


adamhathcock:
> Minimal change, thought I might as well. Great progress recently. Nice one.

Thanks!

Gotta quiz why these frameworks should be supported.  The dotnet ones are out of support and netstandard 2.1 seems redundant now.  I get there's little to do but still work to support.

I'd not do 4.8 but had requests and netstandard 2.0 seemed easy to do.

Nanook:
People still run older (out of support) operating systems too. I always think it's worth supporting them if it's easy, especially with such a key 100% managed package. I understand where you're coming from though. It's fine if you don't want to merge it too :-). I just noticed the code was already 99% compliant.

adamhathcock:
I'll support it for now.  I just don't want things to get in my way more than they already are!  Though, I suppose supporting net4.8 or net standard 2.0 is the worst

Nanook:
Having NetStandardX.X covers most legacy requirements to be fair. I will admit to getting carried away :).

PR #1266, opened 2026-03-27
Fix three BLAKE2sp correctness bugs and eliminate allocations in hot path

coderb:
1. Fix off-by-one in Update(BLAKE2SP): pos was masked with 448 (64×7) instead of 511 (64×8−1), causing incorrect leaf assignment when update chunks are not multiples of 64 bytes. This produced wrong hashes whenever streaming reads didn't align to 64-byte boundaries.

2. Fix double-finalization when Read() is called again after returning 0. Final() was called unconditionally on each EOF read, re-running compression on an already-finalized state and corrupting the hash. EnsureHash() guards with a null-check and is idempotent; both sync and async Read paths share the fix. _blake2sp is set to null after finalization so any erroneous post-final calls fail fast rather than silently corrupting state.

3. Fix false-positive CRC check when stream is not fully drained. _hash was initialized to fileHeader.FileCrc in the constructor, so GetCrc() returned the expected CRC rather than the computed one if the stream was abandoned early. The check would then compare the expected hash against itself and always succeed, silently accepting a corrupt or incomplete file. _hash is now null until the stream is fully read; GetCrc() throws if called early rather than returning a misleading value.

Additional changes:
- Compress: stackalloc m/v arrays instead of heap; LE fast path via MemoryMarshal.Cast replaces 16 BitConverter.ToUInt32 calls per block
- Final(BLAKE2S): write directly to Span<byte>, eliminating MemoryStream and BitConverter.GetBytes allocations; 8 leaf digests now stackalloc'd
- Inner classes and fields: private, sealed, readonly where applicable


coderb:
adam- i think maybe the repo has an issue that could be affecting my pull req:

IArchiveExtensions.cs(36,22): error IDE0051: Private member 'IArchiveExtensions.extension(IArchive).WriteToDirectoryInternal' is unused

adamhathcock:
Looks like I need to fix master.  I'm pretty I did a fix for release.

Thanks for this.  I'll push things and merge later today.

PR #1279, opened 2026-04-10
Fix setting invalid access time fails extraction

aromaa:
Ignore exceptions when setting the access time as the time might be invalid or the OS rejected it and we should not fail the extraction because of it.

Did not end up writing tests for this as this is region / time / historical data / OS specific issue. The archive I was inspecting had dates with leap seconds that did not line up correctly.

### DIFF
--- a/src/SharpCompress/Common/IEntry.Extensions.cs
+++ b/src/SharpCompress/Common/IEntry.Extensions.cs
@@ -23,17 +23,38 @@ internal static class EntryExtensions
             {
                 if (entry.CreatedTime.HasValue)
                 {
-                    nf.CreationTime = entry.CreatedTime.Value;
+                    try
+                    {
+                        nf.CreationTime = entry.CreatedTime.Value;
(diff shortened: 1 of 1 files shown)


adamhathcock:
I don't understand what exception is happening here.  It's not a nullable issue, right?

aromaa:
The underlying sys call fails and a `ArgumentOutOfRangeException: 'Not a valid Win32 FileTime.'` is thrown.

This is also what the `System.IO.Compression` does https://github.com/dotnet/runtime/blob/29a4e371b0ea3aacb4b44562c7919af76bb1b8f3/src/libraries/Common/src/System/IO/Archiving.Utils.cs#L35

Testing the same archive with `ZipFile.ExtractToDirectory` also does in fact reach the catch, they just suppress the exception unconditionally.

adamhathcock:
Fair enough, just allergic to those type of catches.

Thanks!

PR #1293, opened 2026-04-20
fix: Change LeaveStreamOpen default from true to false

puk06:
As of v0.21, the library is documented to close wrapped streams by default. However, the ReaderOptions.LeaveStreamOpen property defaulted to true, which contradicted the documented behavior and caused file locks when deleting archives after extraction.

Changes:
- Set LeaveStreamOpen = false as the default (consistent with v0.21 design)
- Updated ReaderOptions.cs with comprehensive documentation explaining:
  * File-based vs caller-provided stream handling
  * When to use LeaveStreamOpen = true
  * Usage examples for both scenarios
- Updated AGENTS.md Stream Handling Rules section with:
  * Clear explanation of default disposal behavior
  * Overload differences (file-based vs stream-based)
  * Guidance on using ForExternalStream preset

This ensures the implementation matches the documented behavior and prevents unexpected file locking issues during archive deletion.

Fixes: File deletion failures after archive extraction

---

Please note that while this change aligns with the documented specification, it is a breaking change that may affect existing code.
The removal of the explicit `LeaveStreamOpen=false` override in `ZipArchive` was motivated by the fact that `false` is now the default. However, since other Archive types originally relied on the default value of `true`, existing code that does not explicitly specify ReaderOptions may behave differently than before.
Breaking changes of this nature should be documented somewhere. That said, since I'm not familiar enough with where exactly the documentation should be updated, I've opted not to include those changes in this pull request.

### DIFF
--- a/src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
+++ b/src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
@@ -57,7 +57,7 @@ public partial class ZipArchive
             new SourceStream(
                 files[0],
                 i => i < files.Count ? files[i] : null,
-                readerOptions ?? new ReaderOptions()
+                readerOptions ?? ReaderOptions.ForFilePath
             )
         );
     }
--- a/src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
+++ b/src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
@@ -41,7 +41,7 @@ public partial class ZipArchive
             new SourceStream(
                 fileInfo,
                 i => ZipArchiveVolumeFactory.GetFilePart(i, fileInfo),
-                readerOptions ?? new ReaderOptions()
+                readerOptions ?? ReaderOptions.ForFilePath
             )
         );
     }
--- a/src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
+++ b/src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
@@ -41,7 +41,7 @@ public partial class ZipArchive
             new SourceStream(
                 fileInfo,
                 i => ZipArchiveVolumeFactory.GetFilePart(i, fileInfo),
-                readerOptions ?? new ReaderOptions() { LeaveStreamOpen = false }
+                readerOptions ?? new ReaderOptions()
             )
         );
     }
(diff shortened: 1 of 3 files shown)


puk06:
Please note that while this change aligns with the documented specification, it is a breaking change that may affect existing code.
The removal of the explicit `LeaveStreamOpen=false` override in `ZipArchive` was motivated by the fact that `false` is now the default. However, since other Archive types originally relied on the default value of `true`, existing code that does not explicitly specify ReaderOptions may behave differently than before.
Breaking changes of this nature should be documented somewhere. That said, since I'm not familiar enough with where exactly the documentation should be updated, I've opted not to include those changes in this pull request.

src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
```diff
@@ -57,7 +57,7 @@ public static IWritableArchive<ZipWriterOptions> OpenArchive(
             new SourceStream(
                 files[0],
                 i => i < files.Count ? files[i] : null,
-                readerOptions ?? new ReaderOptions() { LeaveStreamOpen = false }
+                readerOptions ?? ReaderOptions.ForFilePath
```
adamhathcock, line 60: I think this is the answer instead of changing the default

src/SharpCompress/Readers/ReaderOptions.cs
```diff
@@ -24,9 +24,36 @@ namespace SharpCompress.Readers;
...
+    /// using var archive = GZipArchive.OpenArchive(filePath); // LeaveStreamOpen = false
+    ///
+    /// // Caller-provided stream: caller manages lifetime
+    /// using var stream = File.OpenRead(filePath);
+    /// var options = new ReaderOptions { LeaveStreamOpen = true };
+    /// using var archive = GZipArchive.OpenArchive(stream, options);
+    /// </code>
+    /// </para>
+    /// </remarks>
+    public bool LeaveStreamOpen { get; init; } = false;
```
adamhathcock, line 56: I want to keep this as the default I think

adamhathcock (changes requested):
I think the defaults for the FilePath/string overloads need to be changed which look wrong

src/SharpCompress/Archives/Zip/ZipArchive.Factory.cs
puk06: I replaced the section on line 44 in ZipArchive.Factory.cs with `ReaderOptions.ForFilePath`.

puk06:
I believe I’ve fixed the issues you pointed out. Please let me know if there are any other mistakes (I’m not very familiar with GitHub pull requests, so I might have made a mistake somewhere... I’m sorry).

adamhathcock (approved):
Reviewed my comments on 0.21 as you're right.  This will be made consistent with further work.  Thanks for this!

PR #1309, opened 2026-04-29
fix: release source file handles in async archive add entry

puk06:
This pull request refactors how files are added to writable archives, both synchronously and asynchronously, to improve resource management and prevent file locking issues. It introduces buffered streams when adding files, ensuring that the source files can be accessed or deleted immediately after archiving. Additionally, a new asynchronous test verifies that source files are not locked during the archive creation process.

**Resource management and file locking improvements:**

* Refactored `AddEntry` in `IWritableArchiveExtensions.cs` to buffer file contents in memory before adding them to the archive, preventing the archive from holding open file handles on the source files. [[1]](diffhunk://#diff-4ab4fbb9acbd843bb76dcbb65a3e6d752eca4e6b630faa7ff533c5bb13048518L28-R28) [[2]](diffhunk://#diff-4ab4fbb9acbd843bb76dcbb65a3e6d752eca4e6b630faa7ff533c5bb13048518R49-R70)
* Refactored `AddEntryAsync` in `IWritableAsyncArchiveExtensions.cs` to buffer file contents asynchronously, ensuring source files are not locked during or after the async archiving process. [[1]](diffhunk://#diff-afdb0a9380673b9e21cda9070a5239c01a4bfe8c5b851b15d3fdea9a3190262bL31-R31) [[2]](diffhunk://#diff-afdb0a9380673b9e21cda9070a5239c01a4bfe8c5b851b15d3fdea9a3190262bL59-R76)

**Testing for file lock issues:**

* Added a new test `Tar_Random_Write_Add_Releases_Source_File_Async` in `TarArchiveAsyncTests.cs` to verify that source files can be opened exclusively after being added to an archive, confirming that file locks are released as expected.
* Updated test imports in `TarArchiveAsyncTests.cs` to include `System.Runtime.InteropServices` for platform-specific file lock checks.

### DIFF
--- a/src/SharpCompress/Archives/IWritableArchiveExtensions.cs
+++ b/src/SharpCompress/Archives/IWritableArchiveExtensions.cs
@@ -25,13 +25,7 @@ public static class IWritableArchiveExtensions
                 )
                 {
                     var fileInfo = new FileInfo(filePath);
-                    writableArchive.AddEntry(
-                        filePath.Substring(directoryPath.Length),
-                        fileInfo.OpenRead(),
-                        true,
-                        fileInfo.Length,
-                        fileInfo.LastWriteTime
-                    );
+                    writableArchive.AddEntry(filePath.Substring(directoryPath.Length), fileInfo);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


puk06:
This change carries some risk of temporarily consuming a large amount of memory, so I understand if accepting it feels a bit risky. The reason I submitted it as a pull request is that, with the current implementation shown below:

```csharp
private static async Task CreateTarArchive(string sourceFolder, string outputTarFile)
{
    if (!Directory.Exists(sourceFolder)) throw new DirectoryNotFoundException(sourceFolder);
    await using IWritableAsyncArchive<TarWriterOptions> archive = await TarArchive.CreateAsyncArchive();
    foreach (string filePath in EnumerateFiles(sourceFolder))
    {
        string relativePath = Path.GetRelativePath(sourceFolder, filePath);
        await archive.AddEntryAsync(relativePath, filePath);
    }
    using FileStream fileStream = new(outputTarFile, FileMode.Create, FileAccess.Write, FileShare.None, bufferSize: 1024 * 1024, FileOptions.SequentialScan);
    await archive.SaveToAsync(fileStream, new TarWriterOptions(CompressionType.None) { LeaveStreamOpen = false });
}
```

After writing code like this, attempting to delete `sourceFolder` immediately afterward fails indefinitely because the file handles opened during `AddEntryAsync` are never released.

By loading the file contents into memory first, the handles are released promptly, which allows `sourceFolder` to be deleted as expected.

I plan to open an Issue for this separately, but I'm submitting this PR in the meantime as one possible approach. I'd appreciate it if you'd consider merging it.

puk06:
I just confirmed that the same issue occurs with `IWritableArchive` as well, not just `IWritableAsyncArchive`, so I believe the same fix needs to be applied there too.

puk06:
After investigating the issue, I found that the WritableArchive wasn't closing properly, so I've closed this issue and am submitting a new pull request.

PR #1310, opened 2026-04-29
fix: dispose source streams in writable archive async add entry

puk06:
This pull request introduces an important resource management improvement to the archive disposal process and adds a new test to ensure source files are properly released after creating a tar archive asynchronously.

Resource management improvements:

* Added an override for `DisposeAsync` in `AbstractWritableArchive.Async.cs` to ensure that all entries in `newEntries`, `removedEntries`, and `modifiedEntries` are properly closed when the archive is disposed asynchronously. This helps prevent resource leaks and file locks.

Testing enhancements:

* Added a new test `TarAsync_AddEntryAndReleaseSourceFile` in `TarArchiveAsyncTests.cs` to verify that after creating and disposing a tar archive asynchronously, the source file is no longer locked and can be deleted, ensuring correct resource release.

### DIFF
--- a/src/SharpCompress/Archives/AbstractWritableArchive.Async.cs
+++ b/src/SharpCompress/Archives/AbstractWritableArchive.Async.cs
@@ -122,6 +122,14 @@ public abstract partial class AbstractWritableArchive<TEntry, TVolume, TOptions>
             .ConfigureAwait(false);
     }
 
+    public override async ValueTask DisposeAsync()
+    {
+        await base.DisposeAsync().ConfigureAwait(false);
+        newEntries.Cast<Entry>().ForEach(x => x.Close());
+        removedEntries.Cast<Entry>().ForEach(x => x.Close());
+        modifiedEntries.Cast<Entry>().ForEach(x => x.Close());
+    }
+
     protected abstract ValueTask SaveToAsync(
         Stream stream,
         TOptions options,
--- a/tests/SharpCompress.Test/Tar/TarArchiveAsyncTests.cs
+++ b/tests/SharpCompress.Test/Tar/TarArchiveAsyncTests.cs
@@ -177,6 +177,50 @@ public class TarArchiveAsyncTests : ArchiveTests
     }
 
     [Fact]
+    public async ValueTask TarAsync_AddEntryAndReleaseSourceFile()
+    {
+        // Arrange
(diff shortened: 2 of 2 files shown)


adamhathcock:
I've noticed there's a lot of issues around writing and async.  I'm working on a fix now

puk06:
Good luck!
Feel free to use this pull request or close it—either is fine!

adamhathcock:
Closing in favor https://github.com/adamhathcock/sharpcompress/pull/1311

puk06:
I have a question about this previously closed PR.

Was `DisposeAsync` actually included in the 0.49.0 release? Looking at the current master branch, it still does not appear to be implemented. This means that even when using the `await using` pattern as shown below, the resource is never properly released since there is no actual disposal logic:

```csharp
await using IWritableAsyncArchive<TarWriterOptions> archive = await TarArchive.CreateAsyncArchive();
```

Could you clarify whether this was intentionally deferred, or if it was perhaps missed in this PR? Thank you.

adamhathcock:
There is an attempt to do that.  That should be fine.  Please open a new issue if you think it's wrong

I believe I test that here:

https://github.com/adamhathcock/sharpcompress/blob/a984a3e30a1986a7914245c10ab67f86434b683f/tests/SharpCompress.Test/Tar/TarArchiveAsyncTests.cs#L165

puk06:
Thank you very much. There's a chance it may have been my misunderstanding, so I'll test it on my end as well.
Thank you for taking the time to reply.

puk06:
> There is an attempt to do that. That should be fine. Please open a new issue if you think it's wrong
> 
> I believe I test that here:
> 
> https://github.com/adamhathcock/sharpcompress/blob/a984a3e30a1986a7914245c10ab67f86434b683f/tests/SharpCompress.Test/Tar/TarArchiveAsyncTests.cs#L165

The existing Tar_Create_New_Async test is not sufficient to reproduce this issue because it only verifies that the archive contents are correct. It does not attempt to delete the output directory immediately after SaveToAsync, so it cannot detect whether the output file is still locked. To reproduce the bug, the test must try to delete the output directory right after SaveToAsync while the writable archive is still in scope.

So I’ve opened a new issue, so I’d appreciate it if you could refer to that one.

PR #1362, opened 2026-06-30
Zip: fix Zip64 streaming reader corrupting entry size/CRC and failing on non-seekable streams

pathartl:
The Zip64 branch of the streaming header reader assumed a data descriptor always follows a >=4GB entry. When the entry instead has back-patched sizes (no descriptor), the following header was parsed as descriptor fields, overwriting the entry's correct size/CRC with central-directory 0xFFFFFFFF sentinels and, on non-seekable async streams, leaving the reader misaligned so the next entry threw.

Now it detects a header signature after the entry data and leaves the already-correct metadata untouched, parsing that header normally. The fix has been applied to both the sync and async readers.


adamhathcock:
Smells right to me.  Thanks for this!

PR #1369, opened 2026-07-21
Fix zip archive corruption when using async APIs and output stream is non-seekable

Noctree:
**⚠Notice:**
This bug was both analysed and patched by Claude Opus/Fable, including writing the tests,
I have reviewed the code and tested the fix in our app where it was originally discovered and it works.

**Details:**
ZipWriter branches on non-seekable streams by checking the provided stream.CanSeek property, this causes issues with the async writer. Async writer buffers writes inside a MemoryStream, MemoryStream always reports CanSeek == true, even though the actual output stream may not support it. This causes archive corruption when using a non-seekable stream as output (e.g. http response body).

**Fix:**
Capture the destination's CanSeek on ZipCentralDirectoryEntry when the entry is created, and base all seekability decisions in WriteHeader and ZipCentralDirectoryEntry.
Write on that captured value instead of the stream being written into. Sync output is unchanged (there the write target is the real output stream); async output to non-seekable streams is now byte-identical to sync, verified by new regression tests that structurally validate the streamed layout (bit 3 in local + central headers, deferred size fields, trailing descriptors) and assert sync/async byte parity.

**Side effect:**
Async + zip64 on a non-seekable output now throws NotSupportedException at header-write time, matching the sync path, instead of silently proceeding.

### DIFF
--- a/src/SharpCompress/Writers/Zip/ZipCentralDirectoryEntry.cs
+++ b/src/SharpCompress/Writers/Zip/ZipCentralDirectoryEntry.cs
@@ -18,13 +18,15 @@ internal class ZipCentralDirectoryEntry
         ZipCompressionMethod compression,
         string fileName,
         ulong headerOffset,
-        IArchiveEncoding archiveEncoding
+        IArchiveEncoding archiveEncoding,
+        bool outputCanSeek
     )
     {
         this.compression = compression;
         this.fileName = fileName;
         HeaderOffset = headerOffset;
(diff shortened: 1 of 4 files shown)


adamhathcock:
I see the problem.  Thanks for catching it.  Might do further adjustments as well.

adamhathcock:
@Noctree formatting should be ran

adamhathcock:
I made a new PR for this issue.  Thanks!

PR #1381, opened 2026-07-28
Generate the synchronous archive entry extensions from the async ones

virzak:
## Why

`IArchiveEntryExtensions` carries five synchronous overloads written out by hand next to the async ones they mirror. Keeping two copies of the same logic has cost the file some accuracy: the async `WriteToAsync(Stream, ExtractionOptions, ...)` passes `options.BufferSize` down, while the synchronous `WriteTo` of the same shape passed it too, but the private worker differed in how progress was wrapped, and every future change has to be made twice and reviewed twice.

This replaces the hand-written copies with generated ones. The async methods stay as the single source; the synchronous versions are emitted at compile time into the same partial class.

## What it changes for consumers

Nothing. Same type, same method names, same overloads, same signatures, same XML documentation. `dotnet build` on this branch produces the identical public surface, and the full test suite passes unchanged (1888 tests, same as master).

## How

[Zomp.SyncMethodGenerator](https://github.com/zompinc/sync-method-generator) is a Roslyn source generator. An async method marked with `[CreateSyncVersion]` gets a synchronous twin: `await` is unwrapped, `ValueTask` becomes `void`, `CancellationToken` parameters are dropped, `.ConfigureAwait(false)` disappears, `OpenEntryStreamAsync` resolves to `OpenEntryStream`, and so on. `PreserveProgress = true` keeps the `IProgress<T>` parameter, which would otherwise be treated as async-only machinery.

It is a build-time only dependency:

```xml
<PackageReference Include="Zomp.SyncMethodGenerator" PrivateAssets="all" />
```

so it adds no runtime reference and nothing to the shipped package. Consumers of SharpCompress never see it.

Net effect on the source file is -72/+7 lines.

## Two things worth flagging

1. **The doc summaries lose the word "asynchronously."** A summary is now shared by both copies of a method, so it is phrased without tense. If you would rather keep the async wording, the alternative is a per-method summary override, which is more ceremony than it saves here.

1. **The generated code is not pretty.** Rewritten argument lists keep the indentation of the original multi-line layout without the line breaks, so a generated body can read like `WriteEntryToFile(entry,          destinationFileName,`. It compiles correctly and is `// <auto-generated/>`, but it is not what a person would write. I have filed it upstream as [zompinc/sync-method-generator#152](https://github.com/zompinc/sync-method-generator/issues/152) with the analysis. Happy to hold this PR until that is fixed if you would prefer the output be clean first.

## Disclosure

I maintain the generator. I am proposing this because SharpCompress is a good fit rather than to seed a dependency: the async and sync halves here are genuinely mechanical translations of each other, which is exactly the case the tool handles. If you would rather not take a source generator dependency, no hard feelings and I will close this.

To inspect what is generated:

```
dotnet build src/SharpCompress/SharpCompress.csproj -p:EmitCompilerGeneratedFiles=true
```

then look under `obj/`.

## Verification

- `dotnet build SharpCompress.slnx -c Release` - clean, no warnings
- Full test suite on net10.0 - 1888 total, 0 failed, identical to master
- `dotnet csharpier check` - clean
- `packages.lock.json` updated for the new package only; the ILLink entries are untouched

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,6 +19,6 @@
       Version="17.14.15"
     />
     <GlobalPackageReference Include="PolySharp" Version="1.16.0" />
-    <PackageVersion Include="Zomp.SyncMethodGenerator" Version="2.0.40" />
+    <GlobalPackageReference Include="Zomp.SyncMethodGenerator" Version="2.0.40" />
   </ItemGroup>
 </Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,5 +19,6 @@
       Version="17.14.15"
     />
     <GlobalPackageReference Include="PolySharp" Version="1.16.0" />
+    <PackageVersion Include="Zomp.SyncMethodGenerator" Version="2.0.40" />
   </ItemGroup>
 </Project>
--- a/build/packages.lock.json
+++ b/build/packages.lock.json
@@ -52,6 +52,12 @@
         "resolved": "13.0.0",
         "contentHash": "zcCR1pupa1wI1VqBULRiQKeHKKZOuJhi/K+4V5oO+rHJZlaOD53ViFo1c3PavDoMAfSn/FAXGAWpPoF57rwhYg=="
       },
+      "Zomp.SyncMethodGenerator": {
+        "type": "Direct",
+        "requested": "[2.0.40, )",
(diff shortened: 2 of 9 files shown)


adamhathcock:
I really like this idea.  Will run and give feedback.

Maintaining sync and async paths is annoying.

PR #1382, opened 2026-07-28
Update Zomp.SyncMethodGenerator to 2.0.42

virzak:
Two releases since #1381 was merged, both fixing things this repository ran into. Thanks for merging that one - having real generated output to look at is what turned both of these up.

## 2.0.41 - documentation crefs

Documentation is copied into the generated file verbatim, and a cref in it resolves against the file it lands in rather than the file it was written in. The generated file carried none of the source file's using directives, so crefs which relied on one went unresolved and the compiler reported CS1574 for each. The directives are emitted now.

Not visible here because this project does not build with warnings as errors, but it was a build failure elsewhere.

## 2.0.42 - argument list spacing

The generated extension calls in this repository were coming out as

```cs
IArchiveEntryExtensions.WriteTo(archiveEntry,                     streamToWriteTo,
        Constants.BufferSize,
        progress: progress)                ;
```

and now come out as

```cs
IArchiveEntryExtensions.WriteTo(archiveEntry, streamToWriteTo,
        Constants.BufferSize,
        progress: progress);
```

This is the thing I flagged as unfixed when I opened #1381. Three separate causes, all in how the rewritten call was reassembled. One piece is still outstanding - the break which belongs before the first argument, so the receiver and the first argument share a line - tracked at [zompinc/sync-method-generator#152](https://github.com/zompinc/sync-method-generator/issues/152).

## What changed here

Version in `Directory.Packages.props` and the five lock files. Nothing else. The generated code is identical apart from spacing and the added using directives.

## Verification

- `dotnet build SharpCompress.slnx -c Release` - clean
- Full test suite on net10.0 - 1888 total, 0 failed, same as master
- Lock files carry only the version and hash change; the ILLink entries are untouched

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,6 +19,6 @@
       Version="17.14.15"
     />
     <GlobalPackageReference Include="PolySharp" Version="1.16.0" />
-    <GlobalPackageReference Include="Zomp.SyncMethodGenerator" Version="2.0.40" />
+    <GlobalPackageReference Include="Zomp.SyncMethodGenerator" Version="2.0.42" />
   </ItemGroup>
 </Project>
--- a/build/packages.lock.json
+++ b/build/packages.lock.json
@@ -54,9 +54,9 @@
       },
       "Zomp.SyncMethodGenerator": {
         "type": "Direct",
-        "requested": "[2.0.40, )",
-        "resolved": "2.0.40",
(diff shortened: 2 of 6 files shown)


adamhathcock:
Thanks!

PR #1383, opened 2026-07-28
Use more Sync Generator

adamhathcock:
First batches of Zomp.SyncMethodGenerator from @virzak

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -172,9 +172,9 @@ referenced for every project via `Directory.Packages.props`).
   member that did not exist before is a behaviour change, not a deduplication — in particular, a
   generated `Read(Span<byte>)` replaces `Stream`'s default rent-and-copy shim.
 - Deliberately **not** generated: `Dispose`/`DisposeAsync` (on a `Stream` the generated `Dispose()`
-  cannot override the non-virtual `Stream.Dispose()`), `Flush`/`FlushAsync`, `CopyTo`/`CopyToAsync`,
-  `ReadByte`/`WriteByte` (generated without `override`, so they would hide `Stream`'s), and the
-  `Memory`/`ReadOnlyMemory` overloads that have no sync twin.
+  cannot override the non-virtual `Stream.Dispose()`), `ReadByte`/`WriteByte` (generated without
+  `override`, so they would hide `Stream`'s), and the `Memory`/`ReadOnlyMemory` overloads that have no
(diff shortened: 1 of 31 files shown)


virzak:
Thanks for merging #1382 - and sorry about the ILLink churn in it that you had to fix in `9e3a92a`. My SDK is 10.0.302 against your 10.0.301 pin, so a restore rewrites those entries; I'd stripped them out and then let a verification build put them back before committing. Entirely mine, and avoidable.

Since you're on 2.0.42 now, one bit of `docs/SYNC_METHOD_GENERATION.md` can be simplified for batch 3 onwards:

> The generator fully qualifies types and reflows fluent calls. It emits `new global::System.ObjectDisposedException(...)` and splits `BaseStream\n.Read(...)` over lines. Strip `global::` and type qualifiers **both before and after** collapsing whitespace - each form only matches in one of the two passes.

The reflowing was a bug, fixed in 2.0.42 - argument lists now keep the spacing they were written with, so the two-pass normalisation shouldn't be needed. The qualification half of course stays. 2.0.41 also emits the source file's using directives into the generated file, so crefs in copied documentation resolve.

I built this branch and checked it the way your doc prescribes, against `master`:

- every sync method deleted from a `.cs` file has a generated counterpart
- **no new members** - generated overload counts match what `master` declared by hand, per type. Rule 2 is the one I'd expect people to get wrong, and a generated `Read(Span<byte>)` displacing `Stream`'s rent-and-copy shim is a behaviour change rather than a deduplication, so it's good to see it stated and followed
- bodies match apart from qualification, whitespace, and `this.Skip()` -> `StreamExtensions.Skip(this)`
- 1888 tests pass locally on net10.0

For the batches you've flagged as benchmark-sensitive (4, 9, 10, 11): there are `SyncOnly` regions if a generated body needs to diverge, and ZSMGEN004 is reported at compile time when an awaited call has no synchronous equivalent - so those surface as build errors rather than silently becoming sync-over-async. Happy to look at the Rar reader unification when you get there; that one sounds like it wants a design conversation rather than a batch.

adamhathcock:
> For the batches you've flagged as benchmark-sensitive (4, 9, 10, 11): there are `SyncOnly` regions if a generated body needs to diverge, and ZSMGEN004 is reported at compile time when an awaited call has no synchronous equivalent - so those surface as build errors rather than silently becoming sync-over-async. Happy to look at the Rar reader unification when you get there; that one sounds like it wants a design conversation rather than a batch.

These are just generated AI notes.  I've adjusted them but might not do it all like that.

The RarReader unification is something you're welcome to tackle :)

Thanks!

PR #1388, opened 2026-07-29
Improve LZMA decoding for Solid Archives

julianxhokaxhiu:
The main intention of this PR was to bring down the extraction time for 7z solid archives and while trying to do so, I think I improved also some other archive types. I noticed while running tests that on my own machine on top of the latest master, we went down from ~10 seconds to run the entire `SevenZip` test suite to ~3,9s.

Additionally, your performance benchmark [is showing up to 80% improvements for 7zip archives](https://github.com/julianxhokaxhiu/sharpcompress/actions/runs/30411717736) on my last run on top of my own fork ( with these patches ).

The improvements have been made using the performance agent with Claude Sonnet 5. I've personally tested the changes and they seem to bring no regression, at least locally.

Below you can find a detailed summary of what has been done ( generated using Sonnet 5 ).

Any question feel free to ask :)

---

### Summary

This change addresses significant performance regressions when extracting solid 7z archives, along with a related async correctness bug uncovered during the fix. Local test suite runtime for `SharpCompress.Test.SevenZip` dropped from **9.8s to 3.9s**.

### Problem

Extracting individual entries from a solid 7z folder was **O(N²)**: each file's compressed stream was recreated from the start of the folder and re-decoded/skipped forward to reach that entry's offset. For folders with many files, this meant re-decoding the same bytes over and over.

#### 1. Solid-folder decoder stream caching (sync + async)
- `src/SharpCompress/Common/SevenZip/ArchiveDatabase.cs` / `ArchiveDatabase.Async.cs`: added a cached folder-stream overload of `GetFolderStream`/`GetFolderStreamAsync` that reuses the same decoder stream across sequential entries in the same solid folder, only skipping the incremental delta between entries instead of restarting from the folder's beginning. Falls back to a full restart for non-sequential (backward) access.
- `src/SharpCompress/Common/SevenZip/SevenZipFilePart.cs`: updated both sync and async `GetCompressedStream(Async)` to request the cached folder stream with the correct skip position and entry size, sharing the underlying stream via `ReadOnlySubStream(..., leaveOpen: true)`.
- `src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs`: added `Dispose()` override to release the cached decoder stream.

#### 2. Fixed async correctness regression (PPMd)
The initial caching implementation tracked the cache position as the *start* of the current entry rather than accounting for the bytes the caller would actually consume. Since the shared stream is read directly by the caller (not just skipped), this caused a double-skip on the next entry, corrupting the PPMd decode state and throwing `IncompleteArchiveException: Unexpected end of stream`.

**Fix:** pass the entry size into the cached-stream lookup so the cache position is advanced to `skipSize + entrySize` (the expected position after the entry is fully consumed), matching actual stream consumption.

#### 3. Reduced async read overhead in `BufferedSubStream`
- `src/SharpCompress/IO/BufferedSubStream.cs` / `BufferedSubStream.Async.cs`: added a fast synchronous-completion path to `Read`/`ReadAsync` when requested bytes are already available in the in-memory cache, avoiding unnecessary `async` state-machine overhead on the common (already-buffered) path.

#### Benchmark.NET (`SevenZipBenchmarks`)

| Benchmark | Before | After | Δ |
|---|---:|---:|---:|
| `SevenZipLzma2Extract` (sync, mean) | 3.152 ms | 2.901 ms | ~7.9% faster |
| `SevenZipLzma2ExtractAsync` (mean) | 9.950 ms | 9.971 ms | no change (noise-level) |
| Async allocated memory | 540.75 KB | 540.77 KB | unchanged |
| Sync allocated memory | 239.22 KB | 239.22 KB | unchanged |

#### Test suite

| Metric | Result |
|---|---|
| `SharpCompress.Test.SevenZip` wall-clock runtime | 9.8s → 3.9s |
| `SevenZipArchiveTests` + `SevenZipArchiveAsyncTests` (targeted run) | **114 / 114 passed**, 0 failed |
| Full `SharpCompress.Test` project (all archive formats) | **3,682 / 3,682 passed**, 0 failed, 0 skipped (~30.4s) |
| Full solution build (`SharpCompress.csproj`, all TFMs) | Succeeded — .NET 10, .NET 8, .NET 6, .NET Standard 2.1/2.0, .NET Framework 4.8 |

### Investigated but not pursued

Profiling showed the async LZMA2 path (~3.2–3.4x slower than sync) is dominated by fine-grained `async`/`await` overhead inside the range decoder's per-bit/per-byte decode loop (`Decoder.DecodeNormalAsync` ~23% self CPU, `RangeCoder.Decoder.Normalize2Async` ~13% self CPU, `BitDecoder.DecodeAsyncHelper` ~13% self CPU), rather than actual I/O wait time. A read-ahead buffering attempt at the range-decoder level was tried but reverted — LZMA2 interleaves chunk-header reads and range-decoder reads on the same underlying stream, and buffering at that layer stole bytes meant for the next chunk header, breaking decoding (`IncompleteArchiveException`, reproduced on 6 tests including `SevenZipArchive_Solid_WriteToDirectoryAsync_WithProgress`). Properly fixing this requires restructuring the async decode loop to batch-decode against an already-buffered chunk and await only at genuine buffer-fill boundaries — a larger, higher-risk architectural change recommended as a separate follow-up effort with dedicated test coverage.

### Testing

- All 3,682 tests in `SharpCompress.Test` pass.
- All 114 SevenZip-specific tests (sync + async, all codecs including PPMd/LZMA/LZMA2/BZip2, solid/split archives) pass.
- Full solution builds successfully across all target frameworks (.NET 10, .NET 8, .NET 6, .NET Standard 2.1/2.0, .NET Framework 4.8).

### DIFF
--- a/src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
+++ b/src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
@@ -78,6 +78,12 @@ public partial class SevenZipArchive : AbstractArchive<SevenZipArchiveEntry, Sev
         }
     }
 
+    public override void Dispose()
+    {
+        _database?.DisposeCachedFolderStream();
+        base.Dispose();
+    }
+
     private void LoadFactory(Stream stream)
     {
         if (_database is null)
--- a/src/SharpCompress/Common/SevenZip/ArchiveDatabase.Async.cs
+++ b/src/SharpCompress/Common/SevenZip/ArchiveDatabase.Async.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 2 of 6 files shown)


adamhathcock:
The basic approach is sound.  I didn't realize what was happening.  Thinking more and letting AI think too.

Thanks!

PR #1398, opened 2026-08-02
Improve CRC32 by using the slice-by-16 optimization

julianxhokaxhiu:
This PR, discovered while working on #1399, focuses on another micro optimization that we can win around CRC32 computation using a well known technique named "slice-by-16" used by some famous libraries and solutions out there.

The performance gains are around 5x and all the test pass just fine.

Below the report generated by Claude Sonnet 5 ( as well as assisting in the code patch ).

---

## Summary

Speeds up CRC32 checksum computation by ~5.3x using the well-known "slice-by-16" table-driven technique (zlib-ng / Chromium / Intel's slicing-by-8 approach, extended to 16 tables). No public API changes.

## What changed

- `Crc32Stream.CalculateCrc`'s bulk path now processes 16 bytes per iteration using 16 independent, non-dependency-chained table lookups instead of 16 sequential dependent byte-at-a-time lookups.
- `Crc32Stream.InitializeTable` now returns a 4096-entry table (16 tables × 256) instead of 256 entries. The first 256 entries are byte-for-byte identical to the original table, so every existing internal caller (`ChecksumValidationStream`, `GZipChecksumValidationStream`, `LZipStream`, Arj/7Zip writer code) benefits automatically with **zero call-site changes**.
- Falls back to the original byte-at-a-time loop for buffers under 16 bytes and any tail remainder.

## Correctness

143 tests across every CRC-touching format passed, including:
- `ADCTest` (cross-checked against an independent CRC32 implementation on a real file)
- Xz `Crc32Tests`
- Both Zip CRC test suites (72 tests)
- GZip `GZipCrcExtractionTests`
- Both Arj reader test files
- `SevenZipCrcExtractionTests`
- Both SevenZip writer test files

## Performance

Throwaway benchmark on a 512MB random buffer:

| Implementation   | Throughput  |
|-------------------|------------|
| Byte-at-a-time (before) | 623 MB/s |
| Slice-by-16 (after)     | 3,273 MB/s |

**~5.3x faster**, with identical CRC output confirmed byte-for-byte.

## Scope / impact

This affects any extraction path using `ExtractionOptions.CheckCrc = true` / `WriteEntryToFile`, plus GZip/7Zip/Arj/LZip writing and header validation.

### DIFF
--- a/src/SharpCompress/Crypto/Crc32Stream.cs
+++ b/src/SharpCompress/Crypto/Crc32Stream.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Buffers.Binary;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


adamhathcock:
Thanks!

PR #1399, opened 2026-08-03
Add LZMA Parallelism and Fast Decoder for modern dotnet

julianxhokaxhiu:
I've worked on this patch across the weekend, and the main goal was to increase the speed of decoding real workd LZMA archives that are 2GB or bigger. The test file for me was https://modcdn.win/MoguriMain.7z ( Solid Archive, LZMA:24, 2 blocks )

On the existing codebase across multiple tests i was able to extract this file under 1,2 minutes ( or 1:20 seconds ), but to me this wasn't enough. When I was comparing it with `7za` I was feeling something was off, as the same archive on the same PC was able to be extracted in 10 seconds. So I started to look at what we could optimize.

The main goal I was working on was implementing the following based on the existing [7zip](https://github.com/ip7z/7zip) source code:

- Comparing our decoding flow with what 7zip was using ( across the C and ASM code ) and replicate it as much as possible in C#
- Add support for Parallelism by inheriting the same logics used by 7zip

In order to do so of course I had to use extensively Claude Sonnet 5 with a lot of thinking ( I burned 15% of my quota only to work on this, but it was worth it! ), and I'm finally proud to present a final solution which should hopefully make everyone happy.

The gist of this PR therefore is:

- We have a new Fast LzmaDecoder, reimplemented following the C and ASM implementation, ensuring JIT code is squeezed to the max: this allowed us to jump **from 1,2 min to 45s on a single core**
- We have an OPT-IN Parallel decoder for LZMA/LZMA2 archives: this one allows us to bump **from 1,2 min to 7s in extraction time!**

As I was into adding Parallelism I also ensured the interface is there for Read/Write across codecs, even though at the moment is supported only for LZMA/LZMA2. In order also to verify that I added some tests to ensure this doesn't break across iterations.

**Most of this code targets modern .NET, so old frameworks unfortunately won't be able to make use of this** ( unless someone else comes with a PR that somehow manages to unblock it, I didn't focus much on this aspect as I am targeting .NET 8+ anyway on my personal projects ).

Down below you can find a detailed summary of everything I did try to summarize here myself, but if there is any question feel free to ask. Additionally, the baseline you see mentioned as 59s is because I used a test to only read the archive in memory without extracting it into the disk ( as I had to run this multiple times, I had to limit my I/O on the SSD ).

Any feedback as usual is appreciated :)

---

# PR Report: Improve LZMA decoder using 7-Zip logics for modern .NET

## Summary

This change lands two independent performance improvements to SharpCompress's LZMA/LZMA2/7-Zip decode path, both derived from studying real 7-Zip's reference C/ASM decoder implementations:

1. **An opt-in parallel decode path for 7-Zip solid folders** — when a folder is a single, unencrypted LZMA2 stream backed by a real file, and the caller opts in, the folder is decoded across multiple threads instead of one.
2. **A faster single-threaded LZMA decode loop** ("fast path") for all modern target frameworks, using flat prob-table arrays, buffered stream reads, and branchless (CMOV-style) bit decoding instead of the original per-object/virtual-call decode loop.

Both changes are fully backward compatible: the parallel path is **opt-in and defaults to off**, and the fast decode path produces byte-identical output to the legacy path (verified by the full existing test suite plus new dedicated tests).

---

### New public API surface

| File | Change |
|---|---|
| [IParallelismOptions.cs](../src/SharpCompress/Common/Options/IParallelismOptions.cs) | New interface: `bool EnableParallelism { get; set; }`, default `false` |
| [IReaderOptions.cs](../src/SharpCompress/Common/Options/IReaderOptions.cs), [IWriterOptions.cs](../src/SharpCompress/Common/Options/IWriterOptions.cs) | Now extend `IParallelismOptions` |
| [ReaderOptions.cs](../src/SharpCompress/Readers/ReaderOptions.cs) | Implements `EnableParallelism` |
| [ReaderOptionsExtensions.cs](../src/SharpCompress/Readers/ReaderOptionsExtensions.cs) | New `WithEnableParallelism(bool)` fluent extension |
| `WriterOptions.cs` + `ZipWriterOptions.cs` / `TarWriterOptions.cs` / `GZipWriterOptions.cs` / `SevenZipWriterOptions.cs` | Each implements `EnableParallelism` (currently a no-op — **reserved for a future parallel encoder**; no writer implements parallel encoding yet) |
| `WriterOptionsExtensions.cs` | New `WithEnableParallelism(bool)` fluent extension |
| [docs/API.md](API.md) | Documents the new option on both `ReaderOptions` and `WriterOptions` |

```csharp
// Usage
var options = ReaderOptions.ForFilePath.WithEnableParallelism(true);
using var archive = SevenZipArchive.OpenArchive("big-solid.7z", options);
```

### Engine: `Lzma2ParallelDecoder`

New file: [Lzma2ParallelDecoder.cs](../src/SharpCompress/Compressors/LZMA/Lzma2ParallelDecoder.cs) (293 lines, `internal`, format-agnostic).

- `TryScanBlocks(stream, packStart, packSize, unpackSize)` — a header-only pre-scan (no real LZMA decode) that walks an LZMA2 chunk stream and splits it into independently-decodable blocks at dictionary-reset boundaries (`control == 0x01 || control >= 0xE0`), merging small consecutive segments until each accumulated block reaches `MinBlockSize`. Returns `null` on any parse anomaly — a safe signal for callers to fall back to sequential decode.
- `DecodeBlocksParallel(...)` (`#if !LEGACY_DOTNET`, needs `System.IO.RandomAccess` / .NET 6+) runs `Parallel.ForEach` over the scanned blocks, each reading its own packed bytes positionally from a shared `SafeFileHandle` and writing decoded output positionally to another `SafeFileHandle` — no per-thread stream duplication or locking needed since regions never overlap.
- Thresholds are **grounded in 7-Zip's own source** (not invented): `MinBlockSize = 1 << 14` (16KB, from `Lzma2DecMt.c`), `MaxBlockSize = 1 << 28` (256MB, from `Lzma2DecMtProps_Init`'s `outBlockMax`), `MaxThreads = 32` (from `MtDec.h`'s `MTDEC_THREADS_MAX`).

### Integration point: `ArchiveDatabase.GetFolderStream`

[ArchiveDatabase.cs](../src/SharpCompress/Common/SevenZip/ArchiveDatabase.cs) (+125 lines) gates the new path behind `enableParallelism` and a strict eligibility check (`TryGetParallelDecodedFolderStream`), all of which must hold before parallel decode engages:

- `Environment.ProcessorCount > 1` (matches 7-Zip: nothing to gain on a single core)
- Exactly one coder, and it must be `CMethodId.K_LZMA2` (no BCJ/delta filter chains, no encryption — those depend on more than one LZMA2 stream)
- Exactly one pack stream, zero bind pairs
- Non-null/non-empty coder properties
- The input stream must be (or unwrap via `IStreamStack` to) a real `FileStream` (parallel decode needs positional random-access I/O)
- `TryScanBlocks` must find `> 1` independently-decodable block

If any check fails, execution falls through unchanged to the existing sequential decode — so archives that don't fit this exact shape are completely unaffected, even with `EnableParallelism = true`.

On success, the folder is decoded up front into a `FileOptions.DeleteOnClose` temp file and a stream over that temp file is returned. Genuine decode exceptions are rethrown (not swallowed) after cleanup, since a successful scan already validated the shape as safe.

Wired through:
- [SevenZipFilePart.cs](../src/SharpCompress/Common/SevenZip/SevenZipFilePart.cs) — new `enableParallelism` constructor parameter, threaded into `GetFolderStream`/the 6-arg caching overload.
- [SevenZipArchive.cs](../src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs), [SevenZipArchive.Async.cs](../src/SharpCompress/Archives/SevenZip/SevenZipArchive.Async.cs) — pass `ReaderOptions.EnableParallelism` through to both entry construction and `ExtractAllEntries()`'s solid-folder stream.

**Scope boundary:** the async decode path (`GetFolderStreamAsync`) is deliberately left untouched — parallel decode is sync-only for now.

---

## 2. Faster single-threaded LZMA decode ("fast path")

Gated by `#if !LEGACY_DOTNET` (applies to netstandard2.1+/net6.0+/net8.0/net10.0, not net48/netstandard2.0).

| File | Change |
|---|---|
| [LzmaDecoder.Fast.cs](../src/SharpCompress/Compressors/LZMA/LzmaDecoder.Fast.cs) | **New**, 794 lines — `CodeFast`, a from-scratch decode loop using flat `ushort[]` probability tables instead of per-symbol `BitDecoder` struct objects, with range/code state kept in locals for the whole call |
| [LzmaDecoder.cs](../src/SharpCompress/Compressors/LZMA/LzmaDecoder.cs) | `Code(...)` now dispatches to `CodeFast` on modern TFMs; legacy TFMs keep the original loop verbatim |
| [RangeCoder.cs](../src/SharpCompress/Compressors/LZMA/RangeCoder/RangeCoder.cs) | Adds a 64KB `ArrayPool`-backed buffered reader (`FillFastBuffer`) to replace one virtual `Stream.ReadByte()` call per consumed byte |
| [LzOutWindow.cs](../src/SharpCompress/Compressors/LZMA/LZ/LzOutWindow.cs) | Adds `Fast*` accessors so the hot loop can cache dictionary position/buffer as locals instead of repeated field/property access |
| [LzmaStream.cs](../src/SharpCompress/Compressors/LZMA/LzmaStream.cs) | Calls `SetFastLimit(...)` after `Init` for both plain LZMA and LZMA2 chunk decode |

Key techniques ported from 7-Zip's actual x86/x64 ASM decoder (`LzmaDecOpt.asm`), not just its plain-C reference:
- **Branchless (CMOV-equivalent) bit decode** — computes both possible outcomes unconditionally and selects via a bitmask, avoiding the ~50/50, hard-to-predict `if (code < bound)` branch.
- **One-level-ahead child-probability prefetch** for binary-tree decodes (literal/length/pos-slot trees) — both potential child probabilities are loaded before the current bit is even resolved, hiding load latency behind arithmetic instead of stalling after a mispredict.
- Manual unrolling of the literal-tree decode loop specifically (confirmed via profiling to unblock JIT inlining) — other trees were tried and reverted after being confirmed as net regressions.

### Correctness safety net (regression caught and fixed pre-merge)

The new buffered reader could over-read past a logical stream boundary on **unbounded, non-self-clamping streams** (e.g. Zip entries with a data descriptor, where the compressed size isn't known upfront) — bytes belonging to the *next* entry's header could be consumed with no way to give them back. Fixed by:
- Always calling `SetFastLimit` when the compressed size is known.
- When the limit is unknown, only bulk-reading 64KB when the underlying stream is a `BufferedSubStream` (the one type provably self-clamped to its own remaining size); otherwise falling back to exactly 1 byte per refill, matching the legacy byte-by-byte behavior exactly (no risk, only loses buffering benefit in that one edge case).

### Measured impact

On the `7Zip.solid.7z` benchmark (3080 entries, 4.78GB uncompressed, LZMA2, single folder dominates):
- Baseline (pre-existing per-object decode loop): **~57–59s (~80–84 MB/s)**
- Fast path (this change, single-threaded): **~44.8–46s (~103–107 MB/s)** — a real, reproducible **~20–25% single-thread improvement**
- Parallel opt-in path (multi-core, this change): large additional speedup on eligible solid archives, since independent LZMA2 blocks decode concurrently instead of serially (up to `min(blockCount, ProcessorCount, 32)`x)

---

## 3. Tests added

| File | What it covers |
|---|---|
| [LzmaStreamTests.cs](../tests/SharpCompress.Test/Streams/LzmaStreamTests.cs) (+302 lines) | 9 new `Lzma2ParallelDecoder_*` unit tests against hand-built (uncompressed-chunk) LZMA2 streams: small-segment merging, cutting once `MinBlockSize` is reached, non-restart chunks never causing a cut, mismatched/invalid/truncated input → `null`, empty stream → empty list, real compressed multi-chunk parsing, and (`#if !LEGACY_DOTNET`) a full `DecodeBlocksParallel` byte-identical-output check across multiple blocks |
| [SevenZipArchiveTests.cs](../tests/SharpCompress.Test/SevenZip/SevenZipArchiveTests.cs) (+77 lines) | `SevenZipArchive_EnableParallelism_OptInEngagesParallelDecodePath` — writes a real 7z archive (small dictionary forces multiple LZMA2 restart blocks), then asserts via the `SevenZipReader`'s diagnostics hook that the *sequential* path is used by default and the *parallel* (`FileStream`-backed) path is used once `EnableParallelism = true`, with byte-identical output either way |
| [SevenZipArchiveParallelTests.cs](../tests/SharpCompress.Test/SevenZip/SevenZipArchiveParallelTests.cs) (**new**, +310 lines) | 37 parallel-enabled counterparts of the existing 7z read/extraction test suite (LZMA, LZMA2, LZMA/LZMA2+AES, PPMd, BZip2, Copy, ZSTD incl. split volumes, all BCJ/delta/filter variants, solid contiguous extraction, empty-stream entries, Tar-in-7z) — every test opts in via `ReaderOptions.WithEnableParallelism(true)` to confirm decoded content is identical regardless of which decode path is taken |

**Validation performed:**
- Full solution build (`dotnet build`), all target frameworks.
- Full test suite: **1936 passed, 0 failed** (net10.0).
- `dotnet csharpier check .` — clean across all 1031 files.

---

## Backward compatibility

- `EnableParallelism` defaults to `false` everywhere — **no behavior change for any existing caller** unless explicitly opted in.
- The fast decode path is purely an internal implementation swap on modern TFMs; output is byte-identical to the legacy path (verified by the entire pre-existing test suite, which exercises both LZMA and LZMA2 across Zip, 7-Zip, Xz, and standalone `LzmaStream` scenarios).
- `net48`/`netstandard2.0`/`netstandard2.1` (`LEGACY_DOTNET`) are entirely unaffected — both the fast decode path and the parallel decode path are compiled out (`#if !LEGACY_DOTNET`), those targets keep the original decode loop verbatim.
- No public method signatures were removed or changed in an incompatible way; all new members are additive.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -332,6 +332,10 @@ var hinted = ReaderOptions.ForExternalStream.WithExtensionHint("tar.gz");
 // Increase for non-seekable streams with large detection probes, such as SFX RAR
 var buffered = ReaderOptions.ForExternalStream.WithRewindableBufferSize(1_048_576);
 
+// Opt in to formats' optional parallel decode (e.g. 7-Zip's automatic parallel LZMA2
+// solid-folder decode); default is sequential decoding
+var parallel = ReaderOptions.ForExternalStream.WithEnableParallelism(true);
+
 // Extraction presets
 var safeOptions = ExtractionOptions.SafeExtract;  // No overwrite
 var flatOptions = ExtractionOptions.FlatExtract;  // No directory structure
@@ -362,6 +366,7 @@ var options = new ReaderOptions
     DisableCheckIncomplete = false,
     BufferSize = 81920,
     RewindableBufferSize = 1_048_576,
+    EnableParallelism = false,
 };
 
 var extractionOptions = new ExtractionOptions
@@ -443,6 +448,8 @@ var options = new WriterOptions(CompressionType.Deflate)
 };
 archive.SaveTo("output.zip", options);
 ```
+
+`WriterOptions.EnableParallelism` (also available on `ZipWriterOptions`, `TarWriterOptions`, `GZipWriterOptions`, and `SevenZipWriterOptions`) opts in to a format's optional parallel encode. No writer currently implements parallel encoding, so this is a no-op today; it is reserved for future use and mirrors `ReaderOptions.EnableParallelism`.
 
 ### Extraction behavior
 
--- a/src/SharpCompress/Archives/SevenZip/SevenZipArchive.Async.cs
+++ b/src/SharpCompress/Archives/SevenZip/SevenZipArchive.Async.cs
@@ -46,7 +46,14 @@ public partial class SevenZipArchive
             var file = _database._files[i];
             entries[i] = new SevenZipArchiveEntry(
                 this,
-                new SevenZipFilePart(stream, _database, i, file, ReaderOptions.ArchiveEncoding),
+                new SevenZipFilePart(
+                    stream,
+                    _database,
+                    i,
+                    file,
+                    ReaderOptions.ArchiveEncoding,
+                    ReaderOptions.EnableParallelism
+                ),
                 ReaderOptions
             );
         }
--- a/src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
+++ b/src/SharpCompress/Archives/SevenZip/SevenZipArchive.cs
@@ -54,7 +54,8 @@ public partial class SevenZipArchive : AbstractArchive<SevenZipArchiveEntry, Sev
                         _database,
                         i,
                         file,
-                        ReaderOptions.ArchiveEncoding
+                        ReaderOptions.ArchiveEncoding,
+                        ReaderOptions.EnableParallelism
                     ),
                     ReaderOptions
                 );
@@ -190,7 +191,8 @@ public partial class SevenZipArchive : AbstractArchive<SevenZipArchiveEntry, Sev
                 _currentFolderStream = _archive._database!.GetFolderStream(
                     _archive.Volumes.Single().Stream,
                     folder!,
-                    _archive._database.PasswordProvider
+                    _archive._database.PasswordProvider,
+                    Options.EnableParallelism
                 );
             }
 
--- a/src/SharpCompress/Common/Options/IParallelismOptions.cs
+++ b/src/SharpCompress/Common/Options/IParallelismOptions.cs
@@ -1,0 +1,14 @@
+namespace SharpCompress.Common.Options;
+
+/// <summary>
+/// Options for controlling whether SharpCompress may use multi-threaded processing.
+/// </summary>
+public interface IParallelismOptions
+{
+    /// <summary>
+    /// When true, opts in to any optional parallel (multi-threaded) processing a compressor may
+    /// offer. Compressors that do not implement parallel processing ignore this setting. Default
+    /// is false (single-threaded processing), so behavior is unchanged unless explicitly enabled.
+    /// </summary>
+    bool EnableParallelism { get; set; }
+}
--- a/src/SharpCompress/Common/Options/IReaderOptions.cs
+++ b/src/SharpCompress/Common/Options/IReaderOptions.cs
@@ -3,7 +3,11 @@ using SharpCompress.Providers;
 
 namespace SharpCompress.Common.Options;
 
-public interface IReaderOptions : IStreamOptions, IEncodingOptions, IProgressOptions
+public interface IReaderOptions
+    : IStreamOptions,
+        IEncodingOptions,
+        IProgressOptions,
+        IParallelismOptions
 {
     /// <summary>
     /// Look for RarArchive (Check for self-extracting archives or cases where RarArchive isn't at the start of the file)
--- a/src/SharpCompress/Common/Options/IWriterOptions.cs
+++ b/src/SharpCompress/Common/Options/IWriterOptions.cs
@@ -7,7 +7,11 @@ namespace SharpCompress.Common.Options;
 /// <summary>
 /// Options for configuring writer behavior when creating archives.
 /// </summary>
-public interface IWriterOptions : IStreamOptions, IEncodingOptions, IProgressOptions
+public interface IWriterOptions
+    : IStreamOptions,
+        IEncodingOptions,
+        IProgressOptions,
+        IParallelismOptions
 {
     /// <summary>
     /// The compression type to use for the archive.
--- a/src/SharpCompress/Common/SevenZip/ArchiveDatabase.cs
+++ b/src/SharpCompress/Common/SevenZip/ArchiveDatabase.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using SharpCompress.Compressors.LZMA;
 using SharpCompress.Compressors.LZMA.Utilities;
+using SharpCompress.IO;
 
 namespace SharpCompress.Common.SevenZip;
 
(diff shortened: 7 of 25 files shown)


adamhathcock:
Thanks for this.  I was thinking of going back to the 7z source to see if there's any thing to mine there.

I like that the interfaces are similar to what I was doing for my own attempts: https://github.com/adamhathcock/sharpcompress/issues/1001

There are some things to do like memory management but I like the shape of it (the LZMA internals are more of a mystery to me)

julianxhokaxhiu:
Glad you appreciated the PR. I worked extensively on this during the weekend and I'm really proud of what I was able to make AI come up with. I also had a though of always going to dig there, but where AI shines is pattern magic and the newest models are really nailing it honestly.

Regarding memory management that will be my next item in the todo as I wanted to optimize also the Tar job that shows up in the batch and I think there are some potential degrees of optimization :) I'll keep you posted.

Cheers and thanks again!

eXpl0it3r:
Really looking forward to the next release that includes these improvements! 🐇💨

PR #1403, opened 2026-08-03
Tar: Fix memory leak regression

julianxhokaxhiu:
As promised in this PR I fixed the memory leak that was happening for Tar archives in the benchmark for `Tar: Extract all entries (Reader API)`.

The root cause seems to have been a regression introduced in one of the recent commits and the detailed explanation is written below, however I verified [benchmark results](https://github.com/julianxhokaxhiu/sharpcompress/actions/runs/30855223974) and now the Tar entry is green again with an impressive ** -75.4%** result for memory :)

Any question feel free to ask. This patch was assisted using **Claude Sonnet 5**

---

## Summary

Fixes a memory regression in the Tar `IReaderFactory` sync path where opening a Tar reader via `ReaderFactory.OpenReader` allocated a large (~900 KB–1 MB) rewind ring buffer on every call, even when the underlying stream was already seekable and didn't need one.

This was found via the `Tar: Extract all entries (Reader API)` BenchmarkDotNet benchmark, which regressed from **187.89 μs / 341.24 KB allocated** to **362.72 μs / 1109.79 KB allocated** (~3.25x memory, ~2x time). The `Archive` API path (`TarArchive.OpenArchive`) was unaffected.

## Root cause

`TarFactory.OpenReader` (sync) unconditionally did:

```csharp
var sharpCompressStream = new SharpCompressStream(stream);
sharpCompressStream.StartRecording(TarWrapper.MaximumRewindBufferSize);
```

`TarWrapper.MaximumRewindBufferSize` is ~900,000 bytes, driven by BZip2's rewind requirement for compressed-Tar-wrapper auto-detection (`TarWrapper.Wrappers`, added as part of the PAX/wrapper-detection support). The raw `SharpCompressStream` constructor never enables the existing seekable fast path, so `StartRecording` always allocated a full `RingBuffer` of that size — even for a plain seekable `MemoryStream`/`FileStream`, where a native `Seek()` (via `SeekableSharpCompressStream`) would be sufficient with zero extra allocation.

## Fix

In `src/SharpCompress/Factories/TarFactory.cs`:

1. **Use the seekable fast path when genuinely possible.** Added `CreateNestedRecordingStream(Stream)`, which wraps the stream in a `SeekableSharpCompressStream` (no ring buffer, delegates directly to the underlying stream's native `Seek`) when the stream is truly seekable, falling back to the ring-buffered `SharpCompressStream` otherwise.

2. **Correctly detect true seekability.** `SharpCompressStream.CanSeek` always returns `true` for non-passthrough (ring-buffered) instances — it simulates seekability via its ring buffer regardless of whether the real underlying stream supports seeking. A naive `stream.CanSeek` check is therefore unreliable when `stream` is already a `SharpCompressStream` (as it is here, coming from the outer `ReaderFactory`). Added `IsGenuinelySeekable(Stream)`, which unwraps via `BaseStream()` in a loop to find the true underlying stream before checking `.CanSeek`.

3. **Fixed a latent rewind bug, exposed by the fast path.** The sync `OpenReader` was missing a `Rewind()` call before `StopRecording()` that the async `OpenAsyncReader` already had. With the old always-ring-buffered approach this was silently tolerated (ring buffer replay semantics masked it), but with the seekable fast path — where position tracking is 1:1 with the real stream — skipping the rewind left the stream positioned past the first Tar header, causing `IncompleteArchiveException: Unexpected EOF reading tar file`. Added the missing `Rewind()` to match the async method.

4. **Scoped to the sync path only.** The optimization was intentionally applied only to `OpenReader` (sync), not `OpenAsyncReader`. One test (`Tar_LZip_Writer_Async`) uses an `AsyncOnlyStream` mock that is seekable but throws on synchronous `Read`, to guarantee async code paths never perform sync I/O. `LZipStream`'s constructor performs a synchronous `Read` internally even when created via `CreateAsync` (pre-existing, unrelated behavior). The old ring-buffered stream masked this because the bytes were already served from the ring buffer (recorded during the earlier async `IsMatchAsync` probing); the seekable fast path bypasses buffering entirely and would call the underlying stream's `Read` directly, breaking that test. Since the regressed benchmark exercises the sync `ReaderFactory.OpenReader`, this fully addresses the regression without touching the async path.

## Validation

- Full test suite: **1942/1942 passing on net10.0**, **1845/1845 passing on net48** (zero failures, no filter).
- Diagnostic measurement (throwaway console app referencing `SharpCompress.csproj`, using `GC.GetAllocatedBytesForCurrentThread`): allocation for opening + fully reading a 5-entry in-memory Tar via `ReaderFactory.OpenReader` dropped from **~1,119,024 bytes** to **83,856 bytes** (~13x reduction), with correct entry count and content in both cases.
- Formatted with `dotnet csharpier format .` and verified with `dotnet csharpier check .`.

### DIFF
--- a/src/SharpCompress/Factories/TarFactory.cs
+++ b/src/SharpCompress/Factories/TarFactory.cs
@@ -322,10 +322,21 @@ public class TarFactory
     /// never allocates a rewind ring buffer. For non-seekable streams, a ring-buffered
     /// <see cref="SharpCompressStream"/> is required to support rewinding during format detection.
     /// </remarks>
-    private static SharpCompressStream CreateNestedRecordingStream(Stream stream) =>
-        IsGenuinelySeekable(stream)
-            ? new SeekableSharpCompressStream(stream, leaveStreamOpen: false)
-            : new SharpCompressStream(stream);
+private static SharpCompressStream CreateNestedRecordingStream(Stream stream)
+{
+    if (!IsGenuinelySeekable(stream))
+    {
+        return new SharpCompressStream(stream);
+    }
+
+    // Unwrap buffered SharpCompressStream wrappers so we use the real stream's native Seek.
+    while (stream is SharpCompressStream scs && !scs.IsPassthrough)
+    {
+        stream = scs.BaseStream();
+    }
+
+    return new SeekableSharpCompressStream(stream, leaveStreamOpen: false);
+}
 
     /// <inheritdoc/>
     public IReader OpenReader(Stream stream, ReaderOptions? options)
--- a/src/SharpCompress/Factories/TarFactory.cs
+++ b/src/SharpCompress/Factories/TarFactory.cs
@@ -296,11 +296,42 @@ public class TarFactory
 
     #region IReaderFactory
 
+    /// <summary>
+    /// Checks whether <paramref name="stream"/> is backed by a genuinely seekable stream.
+    /// Note: <see cref="SharpCompressStream.CanSeek"/> always reports <see langword="true"/> for
+    /// non-passthrough instances (it simulates seeking via its ring buffer even over a
+    /// non-seekable underlying stream), so it must be unwrapped to check the real stream.
+    /// </summary>
+    private static bool IsGenuinelySeekable(Stream stream)
+    {
+        while (stream is SharpCompressStream scs)
+        {
+            stream = scs.BaseStream();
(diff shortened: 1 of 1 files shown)


adamhathcock:
This got merged with #1404 

Thanks for this!

